PR #50, opened 2019-11-28
Add keep command

sttts:
A kept cluster is kept for another hour. The keep command can only be issued in the last 30 min. 15 min before expiry the user is notified.

**Totally untested**.

### DIFF
--- a/manager.go
+++ b/manager.go
@@ -16,6 +16,7 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog"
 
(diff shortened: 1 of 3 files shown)


p0lyn0mial:
it would be great to have a command like that!

smarterclayton:
Unfortunately this won’t work because we have multiple places that forcibly terminate the job.  It’s not currently possible to delay shutdown trivially, we have to change multiple places.

I understand how desirable something that lasts longer would be.  It would probably be easier to have a daily cluster corresponding to morning Brno to afternoon US time.  To do that we probably need something like the endurance clusters, because the current launch mechanism is only really useful with 2-3hr jobs at most.

sttts:
> It would probably be easier to have a daily cluster corresponding to morning Brno to afternoon US time.

We have that already by tiredless work from @sallyom to start them every second day. But this only serves very limited use-cases for us as those must not be damaged. When debugging long-running, possibly destructive things a team cluster won't help.

What is missing to get this keep command axactly? Is it prow?

PR #55, opened 2020-01-16
manager: allow clusters to exist for 8h

stevekuznetsov:
Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>


stevekuznetsov:
@smarterclayton PTAL

smarterclayton:
As we discussed in chat, we were going to (IIRC):

1. bump the activeDeadlineSeconds in the template longer (this is our backstop for teardown protection)
2. bump the timeout here slightly higher (4h?)
3. bump the quota slightly higher (20)

And then assess

PR #78, opened 2020-04-09
Lookup releases in "origin" namespace if release contains "okd"

vrutkovs:
This would enable testing OKD releases which may add arbitrary release
edges

/cc @smarterclayton

### DIFF
--- a/manager.go
+++ b/manager.go
@@ -45,6 +45,9 @@ const (
 	maxTotalClusters = 23
 )
 
+// namespaces defines a list of namespaces where release imagestreamtags are looked up
+var namespaces = []string{"ocp", "origin"}
+
 // JobRequest keeps information about the request a user made to create
 // a job. This is reconstructable from a ProwJob.
 type JobRequest struct {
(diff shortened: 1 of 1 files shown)


manager.go
```diff
@@ -638,23 +638,31 @@ func (m *jobManager) resolveImageOrVersion(imageOrVersion, defaultImageOrVersion
 		return unresolved, "", nil
 	}
 
-	is, err := m.imageClient.ImageV1().ImageStreams("ocp").Get("release", metav1.GetOptions{})
```
smarterclayton, line 641: I would suggest instead changing to a for loop where you walk over `[]string{"ocp","origin"}` and search each dir.  If no match, exit.  You might want to extract the code below into a function that takes an image stream and then invoke the method in a loop.

smarterclayton: Build the URL based on the image stream name you find (switch statement), move that into it's own method.

vrutkovs: Updated PTAL. Didn't move this to the function yet so that #81 would not cause too many merge conflicts

smarterclayton:
Gotta rebase

PR #85, opened 2020-05-12
Allow configuring a dedicated cluster for prowjobs

alvaroaleman:


alvaroaleman:
/hold
actually, it needs to be reversed, prowjobs use incluster, rest external

PR #86, opened 2020-05-12
Fix the bot

alvaroaleman:
Reverts https://github.com/openshift/ci-chat-bot/pull/78 as it breaks the cluster bot because of some quoting issue.

/cc @vrutkovs

### DIFF
--- a/manager.go
+++ b/manager.go
@@ -45,9 +45,6 @@ const (
 	maxTotalClusters = 23
 )
 
-// namespaces defines a list of namespaces where release imagestreamtags are looked up
-var namespaces = []string{"ocp", "origin"}
-
 // JobRequest keeps information about the request a user made to create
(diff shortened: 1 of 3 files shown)


manager.go
```diff
@@ -625,14 +622,6 @@ func versionForRefs(refs *prowapiv1.Refs) string {
 
 var reMajorMinorVersion = regexp.MustCompile(`^(\d)\.(\d)$`)
 
-func buildPullSpec(namespace, tagName string) string {
-	var delimiter = ":"
-	if strings.HasPrefix(tagName, "sha256:") {
-		delimiter = "@"
-	}
-	return fmt.Sprintf("registry.svc.ci.openshift.org/%q/release%s%s", namespace, delimiter, tagName)
```
stevekuznetsov, line 633: That `%q` should be a `%s` and is the bug

PR #87, opened 2020-05-13
Increase duration to 3 hours

smarterclayton:
Data has shown this would increase total CI use by < 5%.


smarterclayton:
/hold

verifying we launch correctly

PR #100, opened 2020-06-03
slack.go: Note machine-os-content is valid in test upgrade

ashcrow:
When using test upgrade the to parameter may be a valid and reachable `machine-os-content` image.

/cc @darkmuggle

### DIFF
--- a/slack.go
+++ b/slack.go
@@ -162,7 +162,9 @@ func (b *Bot) Start(manager JobManager) error {
 	})
 
 	slack.Command("test upgrade <from> <to> <options>", &slacker.CommandDefinition{
-		Description: fmt.Sprintf("Run the upgrade tests between two release images. The arguments may be a pull spec of a release image or tags from https://openshift-release.svc.ci.openshift.org. You may change the upgrade test by passing `test=NAME` in options with one of %s", strings.Join(codeSlice(supportedUpgradeTests), ", ")),
+		Description: fmt.Sprintf(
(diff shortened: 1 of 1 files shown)


cgwalters:
Hmm it is?  Where did that code land?

ashcrow:
> Hmm it is? Where did that code land?

It's been a while but, if I remember right, it's less that a change was made to allow it so much as giving a full registry happens to work. As an example `parseImageInput` doesn't seem to enforce a simple image name, just that the there isn't any offending elements in the reference.

@darkmuggle do you remember the specifics by chance? I remember you ran a few tests with the syntax which lead to me opening up this PR :smile:

cgwalters (changes requested):
I just tried and it doesn't seem to work:

> me: test upgrade 4.5 registry.svc.ci.openshift.org/rhcos-devel/4.6-rhel83:4.6.0-0.nightly-2020-09-03-115938 registry.svc.ci.openshift.org/rhcos-devel/machine-os-content@sha256:e694e82d9f200161591e32d51574542675f9f8c58a2e429b4661c4c57c2ae4af
> cluster-bot: unrecognized option: registry.svc.ci.openshift.org/rhcos-devel/machine-os-content@sha256:e694e82d9f200161591e32d51574542675f9f8c58a2e429b4661c4c57c2ae4af

ashcrow:
Hrm ok -- I'm going to close this then. Thanks for following up @cgwalters!

PR #115, opened 2020-12-21
Print release image name after a successfull "build"

danwinship:
This should(?) make it so that after a successful "build", it will tell you the name of the release image so you don't have to click through to the logs to find it out.

Not tested because I don't know how to test it


slack.go
```diff
@@ -383,6 +383,12 @@ func (b *Bot) notifyJob(response slacker.ResponseWriter, job *Job) {
 			b.sendKubeconfig(response, job.RequestedChannel, job.Credentials, comment, job.RequestedAt.Format("2006-01-02-150405"))
 		}
 		return
+	case "build":
+		if len(job.URL) > 0 && job.State == prowapiv1.SuccessState {
+			releaseImage := buildPullSpec(namespaceForJob(job.Name), "latest")
+			response.Reply(fmt.Sprintf("job <%s|%s> succeeded; release image is %s", job.URL, job.OriginalMessage, releaseImage))
+			return
+		}
```
bradmwilliams, line 391: Unfortunately, the pull spec that this will produce, references an incorrect location for the release:
```
job build 4.9.0-0.ci-2021-08-03-115951,openshift/sdn#325 succeeded; release image is registry.ci.openshift.org/ci-ln-scfvhzt/release:latest
```
The build log for the above run, is [here:](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1422624602353307648)

The correct pullspec is actually here:
`registry.build01.ci.openshift.org/ci-ln-scfvhzt/release:latest`

For "builds", you could potentially pass `job.BuildCluster` to a new method (buildClusterPullSpec()?) that creates the appropriate pullSpec for the respective Cluster.

bradmwilliams (changes requested):

slack.go
bradmwilliams: If you go this route please add godoc to describe the difference between the two methods.

danwinship: > Unfortunately, the pull spec that this will produce, references an incorrect location for the release:

ah, I think the PR predates the separate build cluster...

> For "builds", you could potentially pass `job.BuildCluster` to a new method (buildClusterPullSpec()?) that creates the appropriate pullSpec for the respective Cluster.

Hm... but if you have to pass `job.BuildCluster` to it yourself then it's not really `buildClusterPullSpec`, it's more like `genericPullSpec`...

I just changed `buildPullSpec` to take a cluster argument and had the existing callers pass a new const `centralCIRegistry` to it... but I can change that if you'd like something different.

(I couldn't decide whether the `build` in `buildPullSpec` is referring to image builds or if it's just a generic verb like `makePullSpec`? Either way, if it's being called from the `"build"` code now maybe the naming is confusing and it should be something else? `imagePullSpec`?)

bradmwilliams: Unfortunately, `job.BuildCluster` is only the name of the cluster that the job runs on (i.e. app.ci, build01, build02, vsphere).  So, it cannot be used directly as a registry name.  Sorry if I caused any confusion there!

If I had to guess, the `build` in `buildPullSpec` is/was used as a verb to build the pull spec from either an image tag name or it's actual SHA value.

It feels to me like a completely separate method (something like `generateBuildClusterPullSpec`?) that returns the pullSpec with the necessary substitutions (for build cluster, and tag name).

danwinship: > Unfortunately, `job.BuildCluster` is only the name of the cluster that the job runs on (i.e. app.ci, build01, build02, vsphere). So, it cannot be used directly as a registry name.

uh... so how do I convert a cluster name to a registry name? Based on https://docs.ci.openshift.org/docs/how-tos/use-registries-in-build-farm/ there's no direct mapping...

(If you're interested in this feature, it seems like it would probably take you less time to fix up the PR yourself than to explain to me how to fix it, particularly given that AFAIK there is no way I can run my own copy of the code to test against...)

PR #116, opened 2021-01-04
Switching image resolution over to the app.ci cluster

bradmwilliams:


manager.go
```diff
@@ -640,7 +640,7 @@ func buildPullSpec(namespace, tagName string) string {
 	if strings.HasPrefix(tagName, "sha256:") {
 		delimiter = "@"
 	}
-	return fmt.Sprintf("registry.svc.ci.openshift.org/%s/release%s%s", namespace, delimiter, tagName)
+	return fmt.Sprintf("registry.ci.openshift.org/%s/release%s%s", namespace, delimiter, tagName)
```
smarterclayton, line 643: This is not safe and we are going to have to remove it.  WHo uses this code?

I would drop this for now.

smarterclayton: Drop from this PR, move to another.

bradmwilliams: Unfortunately, it looks like this logic is indirectly used by a majority of the slack commands.

smarterclayton: Hrm.  I guess we can deal with it for now.

manager.go
```diff
@@ -695,7 +695,7 @@ func (m *jobManager) resolveImageOrVersion(imageOrVersion, defaultImageOrVersion
 		}
 	}
 
-	return "", "", fmt.Errorf("unable to find a release matching %q on https://openshift-release.svc.ci.openshift.org or https://origin-release.svc.ci.openshift.org", imageOrVersion)
+	return "", "", fmt.Errorf("unable to find a release matching %q on https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com or https://origin-release.apps.ci.l2s4.p1.openshiftapps.com", imageOrVersion)
```
smarterclayton, line 698: Why don't we have a cname redirect for this?

bradmwilliams: It looks like the new cluster utilizes the cert-manager and has exposed these 2 ingresses...
Ocp:
https://amd64.ocp.releases.ci.openshift.org/
Origin:
https://amd64.origin.releases.ci.openshift.org/
I can update the URLs accordingly.

smarterclayton: Instead of making tihs change, let's get a DNS openshift-release.ci.openshift.org and origin-release.* created and pointed, then we can change these to that (in a separate PR).  The pull spec change is fine.

You need to change the chat bot to look up the release stream from app.ci, not api.ci, since api.ci is now out of date.

smarterclayton:
Just one commit please.

manager.go
bradmwilliams: https://github.com/openshift/release/pull/14635

main.go
```diff
@@ -80,13 +82,15 @@ func run() error {
 	if err != nil {
 		return fmt.Errorf("unable to create client: %v", err)
 	}
-	imageClient, err := imageclientset.NewForConfig(config)
+	// Config and Client to access release images
+	releaseConfig, err := loadKubeconfigFromFlagOrDefault(opt.ReleaseClusterKubeconfig, prowJobKubeconfig)
```
smarterclayton, line 86: In order to keep backwards compatibility, if opt.ReleaseClusterKubeConfig is empty it should use BuildClusterKubeconfig.

bradmwilliams: Updated code.

main.go
```diff
@@ -80,13 +82,15 @@ func run() error {
 	if err != nil {
 		return fmt.Errorf("unable to create client: %v", err)
 	}
-	imageClient, err := imageclientset.NewForConfig(config)
+	// Config and Client to access release images
+	releaseConfig, err := loadKubeconfigFromFlagOrDefault(opt.ReleaseClusterKubeconfig, config)
```
smarterclayton, line 86: you need to default this value to BuildClusterKubeconfig if empty 

```
releaseKubeconfig := opt.ReleaseClusterKubeconfig
if len(releaseKubeconfig) == 0  {
  releaseKubeconfig = opt.BuildClusterKubeconfig
}
```

so that this has back compat

smarterclayton: Oh, nm.  Got it.

PR #120, opened 2021-01-18
Kubeconfig watcher

bradmwilliams:
This change establishes a fsnotify on all the underlying kubeconfigs and exits once any of them changes. This is needed in order to
cope with ServiceAccount Token rotations.

Ref: https://issues.redhat.com/browse/DPTP-1763

/cc @alvaroaleman


main.go
```diff
@@ -111,3 +120,31 @@ func run() error {
...
+	}
+
+	go func() {
+		for e := range watcher.Events {
+			if e.Op == fsnotify.Chmod {
+				// For some reason we get frequent chmod events from Openshift
+				continue
+			}
+			klog.Infof("event: %s, kubeconfig changed, exiting to make the kubelet restart us so we can pick them up", e.String())
+			interrupts.Terminate()
```
alvaroaleman, line 145: Maybe just directly `os.Exit(0)`, I don't think indirecting over `interrupts` when the package is otherwise unused will add anything useful?

main.go
```diff
@@ -111,3 +120,31 @@ func run() error {
 		time.Sleep(5 * time.Second)
 	}
 }
+
+func setupKubeconfigWatches(o *options) error {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return fmt.Errorf("failed to set up watcher: %w", err)
+	}
+	for _, candidate := range []string{o.BuildClusterKubeconfig, o.ReleaseClusterKubeconfig, "/var/run/secrets/kubernetes.io/serviceaccount/token"} {
```
alvaroaleman, line 129: Watching the token secret is actually useless I realized later, because nothing will ever update a token secret, instead a new one is created but that will never end up in existing pods, since its added via an admission controller on creation.

What instead should be used are projected SA tokens, client-go can then dynamically reload them without restarts:
* https://github.com/openshift/ci-ns-ttl-controller/pull/10/commits/96e68ed0a46dabadf742d5052b52eb13838ff328
* https://github.com/openshift/release/commit/3de19b0a64b2ee75c727ec4d6a901c8a9086f7b6

main.go
bradmwilliams: Makes sense.  I have removed `interrupts` all together.

main.go
bradmwilliams: I have removed monitoring the token secret, but left the references to the 2 kubeconfig files.  My understanding is that they should be automatically updated if/when the tokens change, the pod should terminate, and restart with the updated kubeconfigs loaded.  Correct?

alvaroaleman: Yeah, they will get updated. If the ci-chat bot at some point only uses in-cluster config, that mechanism won't help anymore though.
/lgtm

PR #121, opened 2021-02-05
prow: support UNRESOLVED_CONFIG for workflow jobs

vrutkovs:
Look for CONFIG_SPEC in job definition and fall back to UNRESOLVED_CONFIG if not found.

vsphere launch is using workflows, so we can test it now

### DIFF
--- a/prow.go
+++ b/prow.go
@@ -237,7 +237,10 @@ func (m *jobManager) newJob(job *Job) error {
 	// find the ci-operator config for the job we will run
 	sourceEnv, _, ok := firstEnvVar(pj.Spec.PodSpec, "CONFIG_SPEC")
 	if !ok {
-		return fmt.Errorf("the CONFIG_SPEC for the launch job could not be found in the prow job %s", job.JobName)
+		sourceEnv, _, ok = firstEnvVar(pj.Spec.PodSpec, "UNRESOLVED_CONFIG")
+		if !ok {
+			return fmt.Errorf("UNRESOLVED_CONFIG or CONFIG_SPEC for the launch job could not be found in the prow job %s", job.JobName)
+		}
 	}
 	sourceConfig, srcNamespace, srcName, err := loadJobConfigSpec(m.coreClient, sourceEnv, "ci")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


prow.go
```diff
@@ -304,6 +310,8 @@ func (m *jobManager) newJob(job *Job) error {
 			}
 			args = append(args, fmt.Sprintf(`--namespace=$(NAMESPACE)`))
 
+			script := fmt.Sprintf(scriptTemplate, configEnvVar, configEnvVar, configEnvVar, configEnvVar)
```
smarterclayton, line 313: This makes me uncomfortable because it's such a big change.  Can we cleanly change it in the script without templating?

smarterclayton: Can we just append something to the front of the script that sets the env var name correctly.

vrutkovs: Reworked this to avoid envvar and bash templating.
Script is being assembled from:
* `scriptPre`
* instead of `CONFIG_SPEC_n` / `JOB_SPEC_n` a section of `scriptChild` is being templated
* `scriptPost` is appended

That makes us move `container.Command` part below `job.Inputs` parsing though

prow.go
```diff
@@ -381,18 +366,51 @@ func (m *jobManager) newJob(job *Job) error {
...
+				args = append(args, arg)
+			}
+			args = append(args, fmt.Sprintf(`--namespace=$(NAMESPACE)`))
+
+			// Assemble script from scriptPre / scriptChild / scriptPost
+			script := scriptPre
+			for i := 0; i < len(configSpecs); i++ {
+				configSpec := configSpecs[i]
+				jobSpec := jobSpecs[i]
+				script += fmt.Sprintf(scriptChild, i, jobSpec, configEnvVar, configSpec)
```
smarterclayton, line 397: This doesn't read any easier and it's actually scaring me more.  We don't need to pass config_spec to children, you can always pass CONFIG_SPEC as UNRESOLVED_CONFIG and it works fine.  So just make the script always use UNRESOLVED_CONFIG and convert CONFIG_SPEC to UNRESOLVED_CONFIG in the job definition.

That removes the need to change generation of the script file (you can rename the env var up front if the job definition has CONFIG_SPEC)

vrutkovs: Oh. so UNRESOLVED_CONFIG would be accepted when no CONFIG_SPEC present? Okay, that makes it much easier

prow.go
```diff
@@ -408,7 +411,7 @@ func (m *jobManager) newJob(job *Job) error {
 
 	data, _ := json.MarshalIndent(sourceConfig, "", "  ")
 	klog.V(2).Infof("Found target job config %s/%s:\n%s", srcNamespace, srcName, string(data))
-	prow.OverrideJobEnvVar(&pj.Spec, "CONFIG_SPEC", string(data))
+	prow.OverrideJobEnvVar(&pj.Spec, "UNRESOLVED_CONFIG", string(data))
```
smarterclayton, line 414: You also need to override config spec to set it to nil.

vrutkovs: Done

smarterclayton:
I reverted this, please follow up with brad

bradmwilliams:
I tracked the issue down to the fact that most of the release jobs do not have `UNRESOLVED_CONFIG` set, so your change wipes out `CONFIG_SPEC` and never actually sets `UNRESOLVED_CONFIG` because it doesnt exist.  The call to `OverrideJobEnvVar()` only overwrites the environment variable if it actually exists:
https://github.com/openshift/ci-chat-bot/blob/f45720e86722ae1104f5ae25560f817aca8a1885/pkg/prow/prow.go#L170-L180

PR #122, opened 2021-02-08
Switch to 4.8 as master branch association

dmage:


smarterclayton:
I typically wait a few days for this because until everything is green pepole outside of engineers get a worse experience (I.e. the default stays 4.7 until we GA).  But we could change that / wait less time / split up what changes if necessary.

manager.go
```diff
@@ -686,11 +686,11 @@ func (m *jobManager) resolveImageOrVersion(imageOrVersion, defaultImageOrVersion
 			}
 			return "", "", fmt.Errorf("no stable, official prerelease, or nightly version published yet for %s", imageOrVersion)
 		} else if unresolved == "nightly" {
-			unresolved = "4.7.0-0.nightly"
+			unresolved = "4.8.0-0.nightly"
```
smarterclayton, line 689: Let's leave nightly and ci at 4.7 for the next week or so, and then we can potentially improve this so we can control more granularly

dmage: Ok, let's keep it as is. Updated.

PR #124, opened 2021-02-10
prow.go: use UNRESOLVED_CONFIG instead of CONFIG_SPEC

vrutkovs:
This would enable the bot using workflow-based prowjobs, which require
`UNRESOLVED_CONFIG` instead of `CONFIG_SPEC`.
cluster-bot would now always set `UNRESOLVED_CONFIG` and remove `CONFIG_SPEC`
if found.

Follow-up for https://github.com/openshift/ci-chat-bot/pull/121, the diff is effectively `prow.RemoveJobEnvVar(&pj.Spec, "CONFIG_SPEC")` instead of setting it to an empty string.

/cc @smarterclayton @bradmwilliams


bradmwilliams (approved):
Changes look good to me, but @smarterclayton will ultimately need to approve as well.

PR #125, opened 2021-02-17
Adjustments to cluster-bot to support step based periodics

jupierce:
1. Post-steps do not run if `stopJob` deletes the entire project.
Changed to terminate ci-operator pod, which triggers post-steps to
run and for templates to gracefully teardown.
2. Step based jobs follow a pod/container naming convention different than
the traditional template jobs. Instead trying to reflect this, step based
jobs will monitor a secret in the test namespace to extract
cluster information. Access to this secret must be granted by the
openshift-cluster-bot-rbac registry step.

### DIFF
--- a/manager.go
+++ b/manager.go
@@ -121,7 +121,8 @@ type Job struct {
 	Platform  string
 	JobParams map[string]string
 
-	Mode string
+	TargetType string
+	Mode       string
 
 	Inputs []JobInput
 
@@ -318,6 +319,14 @@ func (m *jobManager) sync() error {
 			RequestedChannel: job.Annotations["ci-chat-bot.openshift.io/channel"],
 			RequestedAt:      job.CreationTimestamp.Time,
 			Architecture:     architecture,
+		}
+
+		// This is a new annotation and there may be a brief window where not all
+		// prowjobs possess the annotation.
+		if targetType, ok := job.Annotations["ci-chat-bot.openshift.io/targetType"]; ok {
+			j.TargetType = targetType
+		} else {
+			j.TargetType = "template"
 		}
 
 		var err error
@@ -1188,24 +1197,26 @@ func (m *jobManager) TerminateJobForUser(user string) (string, error) {
 	if err != nil {
 		return "", err
 	}
+
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
 	klog.Infof("user %q requests job %q to be terminated", user, name)
-	if err := m.stopJob(name); err != nil {
-		klog.Errorf("unable to terminate running cluster %s: %v", name, err)
+	if job, ok := m.jobs[name]; ok {
+		if err := m.stopJob(job); err != nil {
+			klog.Errorf("unable to terminate running cluster %s: %v", name, err)
+		}
+		job.Failure = "deletion requested"
+		job.ExpiresAt = time.Now().Add(15 * time.Minute)
+		job.Complete = true
 	}
 
 	// mark the cluster as failed, clear the request, and allow the user to launch again
-	m.lock.Lock()
-	defer m.lock.Unlock()
 	existing, ok := m.requests[user]
 	if !ok || existing.Name != name {
 		return "", fmt.Errorf("another cluster was launched while trying to stop this cluster")
 	}
 	delete(m.requests, user)
-	if job, ok := m.jobs[name]; ok {
-		job.Failure = "deletion requested"
-		job.ExpiresAt = time.Now().Add(15 * time.Minute)
-		job.Complete = true
-	}
 	return "the cluster was flagged for shutdown, you may now launch another", nil
 }
 
--- a/manager.go
+++ b/manager.go
@@ -363,7 +363,7 @@ func (m *jobManager) sync() error {
 			j.State = prowapiv1.PendingState
 			j.Failure = ""
 
-			if j.Mode == "launch" {
+			if j.Mode == "launch" && (previous != nil && !previous.Complete) {
 				if user := j.RequestedBy; len(user) > 0 {
 					if _, ok := m.requests[user]; !ok {
 						var inputStrings [][]string
--- a/manager.go
+++ b/manager.go
@@ -1198,14 +1198,15 @@ func (m *jobManager) TerminateJobForUser(user string) (string, error) {
 		return "", err
 	}
 
+	if err := m.stopJob(name); err != nil {
+		return "", fmt.Errorf("unable to terminate: %v", err)
+	}
+
 	m.lock.Lock()
 	defer m.lock.Unlock()
 
 	klog.Infof("user %q requests job %q to be terminated", user, name)
 	if job, ok := m.jobs[name]; ok {
-		if err := m.stopJob(job); err != nil {
-			klog.Errorf("unable to terminate running cluster %s: %v", name, err)
-		}
 		job.Failure = "deletion requested"
 		job.ExpiresAt = time.Now().Add(15 * time.Minute)
 		job.Complete = true
--- a/prow.go
+++ b/prow.go
@@ -132,21 +132,39 @@ func (r *URLConfigResolver) Resolve(org, repo, branch, variant string) ([]byte, 
 
 // stopJob triggers graceful cluster teardown. If this method returns nil,
 // it is safe to consider the cluster released.
-func (m *jobManager) stopJob(name string) error {
-	namespace := fmt.Sprintf("ci-ln-%s", namespaceSafeHash(name))
-	if _, err := m.coreClient.CoreV1().Secrets(namespace).Get(context.TODO(), "launch", metav1.GetOptions{}); err == nil {
-		// If launch secret exists, this is a step registry based test implementation (vs template based).
+func (m *jobManager) stopJob(job *Job) error {
+	// TODO: Replace all this logic with a simple deletion of the associated prowjob in the ci namespace
+	// once https://bugzilla.redhat.com/show_bug.cgi?id=1937523 is fixed on the build farms.
+	// For step-based jobs, until this bug is fixed, deleting the prowjob will SIGKILL (instead of SIGINT) the running
+	// test container. This prevents it from sharing key data with the post steps.
+	namespace := fmt.Sprintf("ci-ln-%s", namespaceSafeHash(job.Name))
 
-		//We need to terminate the workload and let post steps complete gracefully for successful cleanup. The "wait" step in the
-		// launch step must be terminated.
-		workloadPodName := "launch-wait"
-		_, err := commandContents(m.coreClient.CoreV1(), m.coreConfig, namespace,  workloadPodName, "test", []string{"pkill", "-9", "-P", "1"})
-		if err != nil && !errors.IsNotFound(err) {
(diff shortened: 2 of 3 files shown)


manager.go
```diff
@@ -360,7 +360,7 @@ func (m *jobManager) sync() error {
 			j.State = prowapiv1.PendingState
 			j.Failure = ""
 
-			if j.Mode == "launch" {
+			if j.Mode == "launch" && (previous != nil && !previous.Complete) {
```
smarterclayton, line 363: Good catch (can you put this in its own commit)?

prow.go
```diff
@@ -50,7 +50,7 @@ var supportedUpgradeTests = []string{"e2e-upgrade", "e2e-upgrade-all", "e2e-upgr
 
 // supportedPlatforms requires a job within the release periodics that can launch a
 // cluster that has the label job-env: platform-name.
-var supportedPlatforms = []string{"aws", "gcp", "azure", "vsphere", "metal"}
+var supportedPlatforms = []string{"aws", "gcp", "azure", "vsphere", "metal", "osd-aws"}
```
smarterclayton, line 53: hrm, i think osd should just be a platform, and you should be able to say `ows,osd` or just `osd` if you want it.

prow.go
```diff
@@ -222,6 +236,9 @@ func (m *jobManager) newJob(job *Job) error {
 		if len(input.Version) == 0 {
 			continue
 		}
+		// OSD needs to resolve the pullspec itself from cluster version, so include it
+		// in the env vars.
+		prow.SetJobEnvVar(&pj.Spec, "CLUSTER_VERSION", input.Version)
```
smarterclayton, line 241: ugh, this is a net new var i wasn't aware of.  I would probably prefer it be named differently - is only OSD using it right now?

jupierce: Yes - It is only for OSD atm. Their API works based on published Cincinnati versions and can't accept an arbitrary sha. I kept the name generic because I thought future providers might have a similar limitation (e.g. cluster pools), but I'm certainly open to making it OSD specific. `OSD_CLUSTER_VERSION` ?

prow.go
jupierce: Perhaps:
```
var supportedPlatforms = []string{"aws", ..., "osd"}
var supportedParameters = []string{"ovn", "proxy", "compact", "fips", ..., "osd-aws"}
```
This would allow a logic branch in the `osd-create-create` step to handle the variants versus a new (nearly identical) periodic if `aws,osd` is a separate item in `supportedPlatforms`.

manager.go
jupierce: Extracted into first commit.

prow.go
jupierce: Current commit has `osd` as a platform and `osd-aws` as a variant (anticipating variants like `osd-gcp` in the future).

prow.go
smarterclayton: OSD_INSTALL_CLUSTER_VERSION for now would be ideal, with a big TODO: "this needs to be reconciled with a longer term strategy".

dedicated needs to work with images eventually.  Can you verify with Jeremy that's just a point in time thing?

prow.go
```diff
@@ -945,14 +1051,23 @@ func findTargetName(spec *corev1.PodSpec) (string, error) {
...
 					value := (&resolvedEnvironment{env: container.Env}).Resolve(value)
 					if len(value) == 0 {
-						return "", fmt.Errorf("bug in resolving %s", value)
+						return "", false, fmt.Errorf("bug in resolving %s", value)
 					}
-					return value, nil
+					targetName = value
 				}
 			}
+			if strings.HasPrefix(arg, "--template") {
```
smarterclayton, line 1059: This feels like a boolean that should be returned from this method or passed in.  Probably passed in.

prow.go
jupierce: The recommendation has been implemented in https://github.com/openshift/ci-chat-bot/pull/131 . References to OSD have been removed from this PR.

prow.go
jupierce: The result of this check is returned as a boolean by the method. To change to "passed in", are you suggesting a method signature change to something like:
```
func findTarget(spec *corev1.PodSpec, isTemplate *bool) (string, error)
     ...
```
and setting `*isTemplate` in the body?

smarterclayton: no, you should be telling findTarget what type of job this is.  You should know up front whether this is template or step, not guess.  Then you should either have two methods (if necessary) or one method.  But the end goal is not accidentally mistaking one job for another because someone changed something.  THe step vs template determination needs to be done one time at the top of the job, not later.

prow.go
```diff
@@ -130,12 +130,26 @@ func (r *URLConfigResolver) Resolve(org, repo, branch, variant string) ([]byte,
 	}
 }
 
-// stopJob triggers namespace deletion, which will cause graceful shutdown of the
-// cluster. If this method returns nil, it is safe to consider the cluster released.
+// stopJob triggers graceful cluster teardown. If this method returns nil,
+// it is safe to consider the cluster released.
 func (m *jobManager) stopJob(name string) error {
```
stevekuznetsov, line 135: In both cases you should just delete the Pod running `ci-operator` in the appropriate `ci` namespace

prow.go
```diff
@@ -557,41 +569,89 @@ func (m *jobManager) waitForJob(job *Job) error {
...
-			return false, errJobCompleted
-		}
-		pod, err := m.coreClient.CoreV1().Pods(namespace).Get(context.TODO(), targetPodName, metav1.GetOptions{})
-		if err != nil {
-			// pod could not be created or we may not have permission yet
-			if !errors.IsNotFound(err) && !errors.IsForbidden(err) {
-				lastErr = err
+		if stepBasedMode {
+			if m.jobIsComplete(job) {
+				// The goal here is to interrupt the installation pod or any other step involved in
```
stevekuznetsov, line 583: Again you should be able to just delete the Pod running `ci-operator`, it propagates down

prow.go
jupierce: Thanks @stevekuznetsov . Verified this worked for both templates and steps and dramatically simplifies things. Added a new commit with the change.

jupierce: After trying this at various points of the install, discovered https://github.com/openshift/release/pull/16552 where data written into SHARED_DIR is not available to the post pod. Will need to track this down before adopting this approach.

prow.go
```diff
@@ -569,41 +584,67 @@ func (m *jobManager) waitForJob(job *Job) error {
 		return fmt.Errorf("unable to check launch status: %v", err)
 	}
 
-	klog.Infof("Job %q waiting for setup container in pod %s/%s to complete", job.Name, namespace, targetPodName)
+	if stepBasedMode {
+		klog.Infof("Job %q waiting for step secret %s/%s to be populated", job.Name, namespace, targetName)
```
smarterclayton, line 588: I don't think you have to split this message - it doesn't add anything

prow.go
```diff
@@ -955,12 +1033,17 @@ func findTargetName(spec *corev1.PodSpec) (string, error) {
...
+					targetName = value
 				}
 			}
 		}
 	}
-	return "", fmt.Errorf("could not find argument --target=X in prow job pod spec to identify target pod name")
+
+	if len(targetName) > 0 {
+		return targetName, nil
+	} else {
```
smarterclayton, line 1044: don't else when the if returns

smarterclayton: but return errors early

jupierce: All changes to the method reverted. They were no longer necessary with the change to determining step mode.

prow.go
jupierce: Fixed with workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1937523: 
- https://github.com/openshift/ci-tools/pull/1805 
- https://github.com/openshift/ci-tools/pull/1823

jupierce:
A new commit added which simplifies the stopJob implementation now that https://bugzilla.redhat.com/show_bug.cgi?id=1937523 is being worked around in ci-operator in production.

manager.go
```diff
@@ -1185,24 +1194,26 @@ func (m *jobManager) TerminateJobForUser(user string) (string, error) {
...
 		return "", err
 	}
+
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
 	klog.Infof("user %q requests job %q to be terminated", user, name)
-	if err := m.stopJob(name); err != nil {
-		klog.Errorf("unable to terminate running cluster %s: %v", name, err)
+	if job, ok := m.jobs[name]; ok {
```
smarterclayton, line 1202: putting stopjob under the lock violates the "no long running operations under the lock".  You probably need to refactor this to:

1. get the current job status under lock, then release
2. stop the job
3. if stopping is successful, take the log and update the job status

for 1 that means moving the "get job, check status" to a method with lock and defer, then here check the value.  you also want to return an error if stopErr errors out so that the user knows stop request failed.  that means then that we'll never "drop the request" until the stop job passes, but that also may mean we need additional logic here to clean up the state (i.e. you'll need to handle "job doesn't exist" and race conditions)

jupierce: ack on the time spent in the lock. I put `stopJob` inside the lock initially because it required the actual `job` metadata to determine if it was stopping a template or step job. Now that those paths share the same logic, `stopJob` does not require the `job` at all -- so I've moved it back to its original location outside the lock..
I did, however, close one race condition (`stopJob` run before prowjob is `Pending`) and return a message back to the user as requested.

prow.go
```diff
@@ -130,14 +130,39 @@ func (r *URLConfigResolver) Resolve(org, repo, branch, variant string) ([]byte,
...
-// stopJob triggers namespace deletion, which will cause graceful shutdown of the
-// cluster. If this method returns nil, it is safe to consider the cluster released.
+// stopJob triggers graceful cluster teardown. If this method returns nil,
+// it is safe to consider the cluster released.
 func (m *jobManager) stopJob(name string) error {
-	namespace := fmt.Sprintf("ci-ln-%s", namespaceSafeHash(name))
-	if err := m.projectClient.ProjectV1().Projects().Delete(context.TODO(), namespace, metav1.DeleteOptions{}); err != nil && !errors.IsNotFound(err) {
+	uns, err := m.prowClient.Namespace(m.prowNamespace).Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		if !errors.IsNotFound(err) {
```
smarterclayton, line 138: ?  This looks wierd.

smarterclayton: Hrm...

jupierce: Thanks for catching that. Removed `!` .

prow.go
```diff
@@ -130,14 +130,39 @@ func (r *URLConfigResolver) Resolve(org, repo, branch, variant string) ([]byte,
...
+			// There may have been an issue creating the prowjob; treat as success
+			return nil
```
smarterclayton, line 140: It still looks wierd - if we can't find the pod, we return "err can't find pod".  If get any other error, we return nil (but what error are you expecting here that would imply "we can go ahead and release it)?  User needs to retry on any error except NotFound.

jupierce: Sorry, I have to validate on my test branch which enables osd & I accidentally did a relevant push there. For real this time, removing `!` .

jupierce:
Testing looks good.
/unhold

PR #128, opened 2021-03-02
Test infra bump/go mod tidy

bradmwilliams:
As a followup to yesterdays test-infra bump, I thought I'd verify that the chat-bot had the correct version as well.  Interestingly enough, it turns out that the chat-bot does not actually require `k8s.io/test-infa`.  This PR contains 2 commits, one is the initial test-infra bump and the second is a `go mod tidy` that cleans house quite a bit.


bradmwilliams:
/close
Superseded by #135

PR #131, opened 2021-03-03
Enable OSD as a supported platform

jupierce:


jupierce:
Underpinning PR still in flux. Will reopen when it lands.

PR #135, opened 2021-03-09
Converting over to k8s.io/test-infra

bradmwilliams:


pkg/prow/prow.go
```diff
@@ -3,6 +3,7 @@ package prow
 import (
 	"bytes"
 	"fmt"
+	"github.com/openshift/ci-chat-bot/pkg/prow/apiv1"
```
stevekuznetsov, line 6: What is this?

stevekuznetsov: Probably easier for someone's overloaded brain not to have this called Prow API since it's not, but I don't care too much

PR #143, opened 2021-03-25
Allowing cluster launch jobs to run on any build cluster

bradmwilliams:
These changes deprecate the `--build-cluster-kubeconfig` flag and provide a new `--build-cluster-kubeconfigs-location`, which is a path to the folder containing all the kubeconfig files for any build cluster.  The chat-bot will interrogate this location for files with the pattern: (build.cluster).kubeconfig.  Where "build.cluster" is intended to match the `cluster` field in the underlying ProwJob definitions (https://github.com/openshift/release/blob/05adfe385d627a02cc8fb92bf3f500f2997fb36f/ci-operator/jobs/openshift/release/openshift-release-infra-periodics.yaml#L3).  For each kubeconfig it finds, the chat-bot will create a Map entry that contains all the relevant configs/clients to communicate to the respective cluster.  This Map will then be used, throughout the chat-bot, to pull the appropriate configs/client for the job that it is currently processing.

### DIFF
--- a/manager.go
+++ b/manager.go
@@ -22,16 +22,12 @@ import (
 
 	"github.com/blang/semver"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/client-go/dynamic"
-	clientset "k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
-
 	imagev1 "github.com/openshift/api/image/v1"
 	"github.com/openshift/ci-chat-bot/pkg/prow"
 	imageclientset "github.com/openshift/client-go/image/clientset/versioned"
-	projectclientset "github.com/openshift/client-go/project/clientset/versioned"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/dynamic"
 	prowapiv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 )
 
@@ -139,6 +135,7 @@ type Job struct {
 	Complete      bool
 
 	Architecture string
+	BuildCluster string
 }
 
 func (j Job) IsComplete() bool {
@@ -156,16 +153,13 @@ type jobManager struct {
 	maxClusters   int
 	maxAge        time.Duration
 
-	prowConfigLoader  prow.ProwConfigLoader
-	prowClient        dynamic.NamespaceableResourceInterface
-	coreClient        clientset.Interface
-	imageClient       imageclientset.Interface
-	targetImageClient imageclientset.Interface
-	projectClient     projectclientset.Interface
-	coreConfig        *rest.Config
-	prowNamespace     string
-	githubURL         string
-	forcePROwner      string
+	prowConfigLoader prow.ProwConfigLoader
+	prowClient       dynamic.NamespaceableResourceInterface
+	imageClient      imageclientset.Interface
+	clusterClients   BuildClusterClientConfigMap
+	prowNamespace    string
+	githubURL        string
+	forcePROwner     string
 
 	configResolver ConfigResolver
 
@@ -185,11 +179,8 @@ func NewJobManager(
 	prowConfigLoader prow.ProwConfigLoader,
 	configResolver ConfigResolver,
 	prowClient dynamic.NamespaceableResourceInterface,
-	coreClient clientset.Interface,
 	imageClient imageclientset.Interface,
-	targetImageClient imageclientset.Interface,
-	projectClient projectclientset.Interface,
-	config *rest.Config,
+	buildClusterClientConfigMap BuildClusterClientConfigMap,
 	githubURL, forcePROwner string,
 ) *jobManager {
 	m := &jobManager{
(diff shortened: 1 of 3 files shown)


prow.go
```diff
@@ -132,9 +132,9 @@ func (r *URLConfigResolver) Resolve(org, repo, branch, variant string) ([]byte,
 
 // stopJob triggers namespace deletion, which will cause graceful shutdown of the
 // cluster. If this method returns nil, it is safe to consider the cluster released.
-func (m *jobManager) stopJob(name string) error {
-	namespace := fmt.Sprintf("ci-ln-%s", namespaceSafeHash(name))
-	if err := m.projectClient.ProjectV1().Projects().Delete(context.TODO(), namespace, metav1.DeleteOptions{}); err != nil && !errors.IsNotFound(err) {
+func (m *jobManager) stopJob(job *JobRequest) error {
+	namespace := fmt.Sprintf("ci-ln-%s", namespaceSafeHash(job.Name))
+	if err := m.buildClusterClientConfigMap[job.BuildCluster].ProjectClient.ProjectV1().Projects().Delete(context.TODO(), namespace, metav1.DeleteOptions{}); err != nil && !errors.IsNotFound(err) {
```
smarterclayton, line 137: job is not threadsafe, you must put these behind threadsafe accessors.

bradmwilliams: I added a mutex to the JobRequest.  It looks like all the other calls appear to be protected by a mutex at a higher level.

bradmwilliams: Scratch my last comment.  I've moved the call to `stopJob` to within the JobManager mutex...
https://github.com/openshift/ci-chat-bot/blob/1301dbfd3d590318012fbe12ab1de9cb2b364c2e/manager.go#L1186-L1187

manager.go
```diff
@@ -71,6 +67,9 @@ type JobRequest struct {
 	JobParams map[string]string
 
 	Architecture string
+	BuildCluster string
+
+	sync.Mutex
```
smarterclayton, line 72: You can't put a lock on the job request, we copy that around.  You have to put the lock on the thing holding the the cached clients and prevent simultaneous writes to the map - the jobManager.  You need to hide accesses to the map under an accessor method that gives you a client and holds the lock while it updates the cache.

bradmwilliams: @smarterclayton I've made the suggested updates.  I'd like to confirm if these couple of instances, of direct calls to `m.requests`, are OK as they are or should they be updated to the use the `requestsClient()` as well:
```
	for _, req := range m.requests {
		if req.RequestedAt.Add(m.maxAge * 2).Before(now) {
			klog.Infof("request %q is expired", req.User)
			m.requestsClient().Delete(req.User)
		}
	}


	klog.Infof("Job sync complete, %d jobs and %d requests", len(m.jobs), len(m.requests))
```

manager.go
```diff
@@ -1318,3 +1320,31 @@ func (m *jobManager) finishJob(name string) {
...
 	delete(m.muJob.running, name)
 }
+
+type requestsClient struct {
+	lock     *sync.Mutex
+	requests *map[string]*JobRequest
+}
+
+func (m *jobManager) requestsClient() *requestsClient {
+	return &requestsClient{lock: &m.requestsLock, requests: &m.requests}
```
smarterclayton, line 1330: jobManager should have a map by build cluster to client:

```
type ClientForCluster struct {
   CoreClient *coreclient.Interface
   ProjectsClient *projectclient.Interface
   ...
}

type jobManager struct {
  ... 
  clientLock sync.Mutex
  clients map[string]ClientsForCluster
}

func (m *jobManager) Clients(cluster string) {
  m.clientLock.Lock()
  defer m.clientLock.Unlock()
  clients, ok := m.clients[cluster]
  if ok {
    return clients
  }
  ... init clients from the configs
  m.clients[cluster] = clients
  return clients
}

```

Then you have to pass the cluster name around in the job request (and serialize to the annotations and back)

manager.go
```diff
@@ -156,16 +154,13 @@ type jobManager struct {
...
-	targetImageClient imageclientset.Interface
-	projectClient     projectclientset.Interface
-	coreConfig        *rest.Config
-	prowNamespace     string
-	githubURL         string
-	forcePROwner      string
+	prowConfigLoader            prow.ProwConfigLoader
+	prowClient                  dynamic.NamespaceableResourceInterface
+	imageClient                 imageclientset.Interface
+	buildClusterClientConfigMap BuildClusterClientConfigMap
```
smarterclayton, line 160: this should just be... `clusterClients`

manager.go
```diff
@@ -1335,3 +1339,31 @@ func (m *jobManager) finishJob(name string) {
 
 	delete(m.muJob.running, name)
 }
+
+type requestsClient struct {
```
smarterclayton, line 1343: I don't know why you have this.  Every place you call it you already had to be holding the manager lock.  So this doesn't protect anything

manager.go
```diff
@@ -1177,35 +1178,38 @@ func (m *jobManager) LaunchJobForUser(req *JobRequest) (string, error) {
...
+	existing, ok := m.requestsClient().Get(user)
 	if !ok || len(existing.Name) == 0 {
-		return "", fmt.Errorf("no cluster has been requested by you")
+		return "", "", fmt.Errorf("no cluster has been requested by you")
+	}
+	if len(existing.BuildCluster) == 0 {
+		return "", "", fmt.Errorf("unable to determine build cluster for your job")
 	}
-	return existing.Name, nil
+	return existing.Name, existing.BuildCluster, nil
```
smarterclayton, line 1196: This is the only place we need to read req.BuildCluster?  You don't need req.BuildCluster because under the lock in this method you can look up whether there is a known job and get buildcluster from it.  Then you only need to have BuildCluster on the job object.

manager.go
```diff
@@ -175,6 +169,8 @@ type jobManager struct {
 	}
 
 	notifierFn JobCallbackFunc
+
+	requestsLock sync.Mutex
```
smarterclayton, line 173: Remove

manager.go
```diff
@@ -430,14 +425,14 @@ func (m *jobManager) sync() error {
 			delete(m.jobs, job.Name)
 		}
 	}
-	for _, req := range m.requests {
+	for _, req := range m.requestsClient().Map() {
```
smarterclayton, line 428: Remove requestsClient completely and just use `m.requests`

smarterclayton:
Pretty close

smarterclayton:
Squash please and I'll merge

bradmwilliams:
> Squash please and I'll merge

Done

PR #159, opened 2021-05-06
Improve the help text for building release images from PRs

bparees:
It's not clear from my quick survey of the code if this command can support other "from" sources besides PRs...   if it can, this PR is incorrect since it modifies the help text to be specifically targeted at how to build a release image based on PRs.

### DIFF
--- a/slack.go
+++ b/slack.go
@@ -283,8 +283,8 @@ func (b *Bot) Start(manager JobManager) error {
 		},
 	})
 
-	slack.Command("build <from>", &slacker.CommandDefinition{
-		Description: "Create a new release image from one or more pull requests. The successful build location will be sent to you when it completes and then preserved for 12 hours.",
+	slack.Command("build <pullrequest>", &slacker.CommandDefinition{
+		Description: "Create a new release image from one or more pull requests. The successful build location will be sent to you when it completes and then preserved for 12 hours.  Example: `build openshift/operator-framework-olm#68,operator-framework/operator-marketplace#396`",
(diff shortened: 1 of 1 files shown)


slack.go
```diff
@@ -283,8 +283,8 @@ func (b *Bot) Start(manager JobManager) error {
 		},
 	})
 
-	slack.Command("build <from>", &slacker.CommandDefinition{
-		Description: "Create a new release image from one or more pull requests. The successful build location will be sent to you when it completes and then preserved for 12 hours.",
+	slack.Command("build <pullrequest>", &slacker.CommandDefinition{
```
bradmwilliams, line 286: You cant change the name of the request parameter without also updating the logic that parses it out (Line 296).

slack.go
```diff
@@ -283,8 +283,8 @@ func (b *Bot) Start(manager JobManager) error {
...
+		Description: "Create a new release image from one or more pull requests. The successful build location will be sent to you when it completes and then preserved for 12 hours.  Example: `<build openshift/operator-framework-olm#68,operator-framework/operator-marketplace#396`",
```
bradmwilliams, line 287: Small nit...
Did you intend to have a `<` in the example?

bradmwilliams (changes requested):

slack.go
bparees: thanks, fixed

slack.go
bparees: nope, fixed thanks

bradmwilliams (approved):
/lgtm

PR #163, opened 2021-05-10
prow: don't panic when cluster not found in clusterClients

vrutkovs:
Check that build cluster config was picked up on start, so that it would not panic when fetching one from `m.clusterClients`


bradmwilliams (commented):
/lgtm

PR #164, opened 2021-05-18
prow: Use getClusterClient to fetch build cluster config

vrutkovs:
`m.clusterClients[job.BuildCluster]` is a repeating pattern, replace it with panic-safe `getClusterClient`. This is useful for local debugging where all cluster kubeconfigs may not be available.

Follow up for https://github.com/openshift/ci-chat-bot/pull/163


bradmwilliams (approved):
/lgtm

PR #165, opened 2021-05-20
Clone repos in build jobs so that ci-operator can read repo config

smarterclayton:
ci-operator added `.ci-operator.yaml` file support, which requires
jobs that build PRs to clone the repo so that config is available.
Set skip_cloning: false, then set a REPO_PATH for each sub build.
The "from" and "to" parts (inputs 0 and 1) have different root
dirs so that we can have two different PRs for the same repo, which
required us to merge a fix to clonerefs.


prow.go
```diff
@@ -1028,10 +1044,13 @@ pids=()
 for var in "${!CONFIG_SPEC_@}"; do
   suffix="${var/CONFIG_SPEC_/}"
   jobvar="JOB_SPEC_$suffix"
+	srcpath="REPO_PATH_$suffix"
+	srcpath="${working_dir}/${!srcpath}"
```
bradmwilliams, line 1048: Was this level of indention intentional?

PR #169, opened 2021-06-04
Checking /tmp/shared for the installer kubeadmin-password

bradmwilliams:
The `metal` team updated their installer job to utilize the
/tmp/shared location for the installer's artifacts (because they
were leaking secrets when everything was placed under the
/tmp/artifacts location). This PR adds a secondary check to
look for the file, in /tmp/shared, if the first check, in
/tmp/artifacts, fails for any reason.

### DIFF
--- a/prow.go
+++ b/prow.go
@@ -801,10 +801,11 @@ func (m *jobManager) waitForJob(job *Job) error {
 		job.PasswordSnippet = strings.TrimSpace(reFixLines.ReplaceAllString(string(logs), "$1"))
 		password, err := commandContents(clusterClient.CoreClient.CoreV1(), clusterClient.CoreConfig, namespace, targetName, "test", []string{"cat", "/tmp/artifacts/installer/auth/kubeadmin-password"})
 		if err != nil {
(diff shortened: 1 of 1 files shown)


prow.go
```diff
@@ -800,6 +800,9 @@ func (m *jobManager) waitForJob(job *Job) error {
 		}
 		job.PasswordSnippet = strings.TrimSpace(reFixLines.ReplaceAllString(string(logs), "$1"))
 		password, err := commandContents(clusterClient.CoreClient.CoreV1(), clusterClient.CoreConfig, namespace, targetName, "test", []string{"cat", "/tmp/artifacts/installer/auth/kubeadmin-password"})
+		if err != nil {
+			password, err = commandContents(clusterClient.CoreClient.CoreV1(), clusterClient.CoreConfig, namespace, targetName, "test", []string{"cat", "/tmp/shared/installer/auth/kubeadmin-password"})
```
smarterclayton, line 804: this returns the wrong error - you shouldn't overwrite err here because you're hiding the error most jobs return (for an error only some jobs would return)

bradmwilliams: I've updated the logging around these failures.

PR #170, opened 2021-06-07
Commandline support for ProwJob generation

bradmwilliams:
The ART team has requested a way to automate their process
of manually firing off upgrade jobs, via the chat-bot, when
new releases are accepted.  We came up with a solution that
involves using the chat-bot, as a command-line tool, to
generate the necessary ProwJob's and then ART will submit
the jobs accordingly.

The majority of this PR is just refactoring the logic to
allow the generated ProwJob payload's to be exported instead
of submitted to the build cluster. I have added support for
all the existing commands, that generate ProwJobs, using the
same logic for either path (slack or commandline).

A sample commandline execution looks like this:
```
./ci-chat-bot --build-cluster-kubeconfigs-location /var/build-cluster-kubeconfigs --prow-config=/etc/config/config.yaml --job-config=/etc/job-config --command 'test upgrade 4.7.0 4.7.14' --output upgrade.json
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -74,7 +74,7 @@ func run() error {
 	pflag.StringVar(&opt.BuildClusterKubeconfigsLocation, "build-cluster-kubeconfigs-location", opt.BuildClusterKubeconfigsLocation, "Path to the location of the Kubeconfigs for the various buildclusters. Default is \"/var/build-cluster-kubeconfigs\".")
 	pflag.StringVar(&opt.ReleaseClusterKubeconfig, "release-cluster-kubeconfig", opt.ReleaseClusterKubeconfig, "Kubeconfig to use for cluster housing the release imagestreams. Defaults to normal kubeconfig if unset.")
 	pflag.StringVar(&opt.WorkflowConfigPath, "workflow-config-path", "", "Path to config file used for workflow commands")
-	pflag.StringVar(&opt.Command, "command", opt.Command, "The command to run in command-line mode")
+	pflag.StringVar(&opt.Command, "command", opt.Command, "The slack command to run directly and then exit")
 	pflag.StringVar(&opt.Output, "output", opt.Output, "The path, of the file, where to write the prow job JSON")
 	opt.prowconfig.AddFlags(emptyFlags)
 	pflag.CommandLine.AddGoFlagSet(emptyFlags)
--- a/main.go
+++ b/main.go
@@ -44,6 +44,8 @@ type options struct {
 	ReleaseClusterKubeconfig        string
 	ConfigResolver                  string
 	WorkflowConfigPath              string
+	Command                         string
+	Output                          string
 }
 
 func main() {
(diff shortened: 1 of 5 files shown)


thegreyd:
@smarterclayton gentle ping :) For context, ART is waiting for this change to fire up upgrade tests when a z-stream release is created (right now the process is manual)

main.go
```diff
@@ -66,7 +68,9 @@ func run() error {
 	pflag.StringVar(&opt.ForcePROwner, "force-pr-owner", opt.ForcePROwner, "Make the supplied user the owner of all PRs for access control purposes.")
 	pflag.StringVar(&ignored, "build-cluster-kubeconfig", ignored, "REMOVED: Kubeconfig to use for buildcluster. Defaults to normal kubeconfig if unset.")
 	pflag.StringVar(&opt.BuildClusterKubeconfigsLocation, "build-cluster-kubeconfigs-location", opt.BuildClusterKubeconfigsLocation, "Path to the location of the Kubeconfigs for the various buildclusters. Default is \"/var/build-cluster-kubeconfigs\".")
-	pflag.StringVar(&opt.ReleaseClusterKubeconfig, "release-cluster-kubeconfig", "", "Kubeconfig to use for cluster housing the release imagestreams. Defaults to normal kubeconfig if unset.")
+	pflag.StringVar(&opt.ReleaseClusterKubeconfig, "release-cluster-kubeconfig", opt.ReleaseClusterKubeconfig, "Kubeconfig to use for cluster housing the release imagestreams. Defaults to normal kubeconfig if unset.")
+	pflag.StringVar(&opt.Command, "command", opt.Command, "The command to run in command-line mode")
```
smarterclayton, line 72: > The slack command to run directly and then exit

main.go
```diff
@@ -66,7 +68,9 @@ func run() error {
...
+	pflag.StringVar(&opt.Output, "output", opt.Output, "The path, of the file, where to write the prow job JSON")
```
smarterclayton, line 73: Why aren't you just outputting this to stdout?  Is there extra info you don't need?

main.go
bradmwilliams: Updated!

main.go
bradmwilliams: If the `--output` option is not specified, then I was writing the resultant prowjob to the screen.  I updated the logic to write directly to stdout instead of using klog.

smarterclayton: The comment was because we probably don't need the `--output` flag at all.  Just output prow jobs to stdout when --command is provided.

prow.go
```diff
@@ -176,19 +176,30 @@ func (m *jobManager) newJob(job *Job) error {
 	if job.IsComplete() && len(job.PasswordSnippet) > 0 {
 		return nil
 	}
+
+	pj, err := m.generateProwJob(job)
```
smarterclayton, line 180: You need to handle this error

slack.go
```diff
@@ -24,60 +38,30 @@ func NewBot(token string) *Bot {
 	}
 }
 
-func (b *Bot) Start(manager JobManager) error {
-	slack := slacker.NewClient(b.token)
-
-	manager.SetNotifier(b.jobResponder(slack))
-
-	slack.DefaultCommand(func(request slacker.Request, response slacker.ResponseWriter) {
+func (b *Bot) initialize(client *slacker.Slacker, manager JobManager) {
```
smarterclayton, line 41: Would recommend you name this `slack` instead of `client` - it's not really a client, it's a server, and it's confusing to call it `client`.

slack.go
```diff
@@ -24,60 +38,30 @@ func NewBot(token string) *Bot {
...
 
-	slack.Command("launch <image_or_version_or_pr> <options>", &slacker.CommandDefinition{
+	client.Command("launch <image_or_version_or_pr> <options>", &slacker.CommandDefinition{
 		Description: fmt.Sprintf(
 			"Launch an OpenShift cluster using a known image, version, or PR. You may omit both arguments. Use `nightly` for the latest OCP build, `ci` for the the latest CI build, provide a version directly from any listed on https://amd64.ocp.releases.ci.openshift.org, a stream name (4.1.0-0.ci, 4.1.0-0.nightly, etc), a major/minor `X.Y` to load the latest stable version for that version (`4.1`), `<org>/<repo>#<pr>` to launch from a PR, or an image for the first argument. Options is a comma-delimited list of variations including platform (%s) and variant (%s).",
 			strings.Join(codeSlice(supportedPlatforms), ", "),
 			strings.Join(codeSlice(supportedParameters), ", "),
 		),
-		Example: "launch openshift/origin#49563 gcp",
+		Example: launchCommandExample,
```
smarterclayton, line 52: Why did you turn these into constants?

slack.go
```diff
@@ -327,15 +219,206 @@ func (b *Bot) Start(manager JobManager) error {
...
+	if !isDirectMessage(channel) {
+		return "", "", "", nil, fmt.Errorf("this command is only accepted via direct message")
+	}
+
+	return user, channel, message, request.Properties(), nil
+}
+
+func generateJobRequest(user, channel, message string, parameters *proper.Properties, example string) (*JobRequest, error) {
+	switch example {
+	case buildCommandExample:
```
smarterclayton, line 245: This is an antipattern - do not do this.

slack.go
```diff
@@ -327,15 +219,206 @@ func (b *Bot) Start(manager JobManager) error {
...
+	for _, cmd := range client.BotCommands() {
+		parameters, isMatch := cmd.Match(command)
+		if !isMatch {
+			continue
+		}
+
+		definition := cmd.Definition()
+
+		switch definition.Example {
+		case buildCommandExample, launchCommandExample, testCommandExample, testCommandUpgradeExample:
```
smarterclayton, line 395: Yeah, do not do this.

slack.go
bradmwilliams: Done!

prow.go
bradmwilliams: Done!

bradmwilliams:
/hold
Need to update logic to handle latest/greatest commands

PR #172, opened 2021-06-11
Handling the termination of the launch command

bradmwilliams:
With the updated logic, that calls a Delete on the ProwJob pod, to terminate a cluster, we need to pass the signal down to the ci-operator to properly terminate the pod (and ultimately delete the underlying cluster).  This fixes the issue we have been having where clusters have been being orphaned when a user issues a `done` command to the chat-bot.

+1 for having the ability to generate the prowjob yaml, without actually starting a Job (#170).

### DIFF
--- a/prow.go
+++ b/prow.go
@@ -369,9 +369,9 @@ func (m *jobManager) newJob(job *Job) error {
 			envPrefix := strings.Join(restoreImageVariableScript, " ")
 			container.Command = []string{"/bin/bash", "-c"}
 			if job.Mode == "build" {
-				container.Command = append(container.Command, fmt.Sprintf("registry_host=%s\n%s\n\n%s\n%s ci-operator $@", registryHost, script, permissionsScript, envPrefix), "")
+				container.Command = append(container.Command, fmt.Sprintf("registry_host=%s\n%s\n\n%s\n%s %s", registryHost, script, permissionsScript, envPrefix, launchClusterScript), "")
 			} else {
-				container.Command = append(container.Command, fmt.Sprintf("registry_host=%s\n%s\n\n%s ci-operator $@", registryHost, script, envPrefix), "")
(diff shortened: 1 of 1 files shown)


prow.go
```diff
@@ -1082,6 +1082,10 @@ chmod ug+x /tmp/bin/oc
 oc policy add-role-to-group system:image-puller -n $(NAMESPACE) system:authenticated
 `
 
+const launchClusterScript = `ci-operator $@ &
```
smarterclayton, line 1085: why isn't this just inlined?  Really no value in splitting it into its own string.

bradmwilliams: At the time I wrote it, I was worried that I might need more logic than I ended up with.  I've in-lined the logic now.

prow.go
```diff
@@ -369,9 +369,9 @@ func (m *jobManager) newJob(job *Job) error {
 			envPrefix := strings.Join(restoreImageVariableScript, " ")
 			container.Command = []string{"/bin/bash", "-c"}
 			if job.Mode == "build" {
-				container.Command = append(container.Command, fmt.Sprintf("registry_host=%s\n%s\n\n%s\n%s ci-operator $@", registryHost, script, permissionsScript, envPrefix), "")
+				container.Command = append(container.Command, fmt.Sprintf("registry_host=%s\n%s\n\n%s\n%s ci-operator $@ &\nwait", registryHost, script, permissionsScript, envPrefix), "")
 			} else {
-				container.Command = append(container.Command, fmt.Sprintf("registry_host=%s\n%s\n\n%s ci-operator $@", registryHost, script, envPrefix), "")
+				container.Command = append(container.Command, fmt.Sprintf("registry_host=%s\n%s\n%s ci-operator $@ &\nwait", registryHost, script, envPrefix), "")
```
smarterclayton, line 374: Why do you have `&\nwait`?  `& wait`.

bradmwilliams: Gone!

PR #175, opened 2021-06-16
Update test-infra

2uasimojo:
...particularly to pick up https://github.com/kubernetes/test-infra/pull/22502 (`skip_if_only_changed` prow job option)


2uasimojo:
/test linter

There's no error in the output. What gives?

2uasimojo:
> There's no error in the output. What gives?

I needed to `go mod tidy`. It would be neat if the test output gave some indication of the specific problem. Yes, there's the generic note at the top, but the test must be failing due to producing a git diff. I'll look into that...

bradmwilliams (approved):
/lgtm

PR #184, opened 2021-07-15
Bump quota to 48 clusters from 36 clusters

smarterclayton:
We are now sitting at the limit, and we do have slack capacity in
a number of environments.


bradmwilliams (approved):
/lgtm

PR #185, opened 2021-07-15
Abort ProwJob instead of removing it

vrutkovs:
Set AbortedState and update pj instead of deleting the object. This would ensure template/workflow objects are being correctly terminated.
Verified that it properly stops workflow-based tests, but I don't have permissions to ensure template-based clusters are stopped as expected.

### DIFF
--- a/prow.go
+++ b/prow.go
@@ -159,10 +159,10 @@ func (m *jobManager) stopJob(name, cluster string) error {
 		return err
 	}
 
-	// Deleting the prowjob pod will gracefully terminate template or step based jobs. In the case of
-	// steps, this includes running post steps.
-	klog.Infof("ProwJob pod for job %q will be deleted", name)
-	return m.clusterClients[cluster].CoreClient.CoreV1().Pods(m.prowNamespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
+	klog.Infof("ProwJob pod for job %q will be aborted", name)
+	pj.Status.State = prowapiv1.AbortedState
+	_, err = m.prowClient.Namespace(m.prowNamespace).Update(context.Background(), prow.ObjectToUnstructured(&pj), metav1.UpdateOptions{})
+	return err
 }
 
 // newJob creates a ProwJob for running the provided job and exits.


prow.go
```diff
@@ -159,10 +159,11 @@ func (m *jobManager) stopJob(name, cluster string) error {
...
 	}
 
-	// Deleting the prowjob pod will gracefully terminate template or step based jobs. In the case of
-	// steps, this includes running post steps.
-	klog.Infof("ProwJob pod for job %q will be deleted", name)
-	return m.clusterClients[cluster].CoreClient.CoreV1().Pods(m.prowNamespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
+	klog.Infof("ProwJob pod for job %q will be aborted", name)
+	pj.SetComplete()
+	pj.Status.State = prowapiv1.AbortedState
+	_, err = m.prowClient.Namespace(m.prowNamespace).Update(context.TODO(), prow.ObjectToUnstructured(&pj), metav1.UpdateOptions{})
```
stevekuznetsov, line 165: Pass a real context?

stevekuznetsov (approved):
/hold

Lack of a context is suspect but otherwise the approach is sound.

prow.go
```diff
@@ -159,10 +159,11 @@ func (m *jobManager) stopJob(name, cluster string) error {
 		return err
 	}
 
-	// Deleting the prowjob pod will gracefully terminate template or step based jobs. In the case of
-	// steps, this includes running post steps.
-	klog.Infof("ProwJob pod for job %q will be deleted", name)
-	return m.clusterClients[cluster].CoreClient.CoreV1().Pods(m.prowNamespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
+	klog.Infof("ProwJob pod for job %q will be aborted", name)
+	pj.SetComplete()
```
alvaroaleman, line 163: This is wrong, it will result in us not terminating the pod as we ignore Completed jobs:
https://github.com/kubernetes/test-infra/blob/01baa4728fcb61d887c858bf4af0f7ca094fd406/prow/plank/reconciler.go#L735-L738

vrutkovs: Removed

prow.go
vrutkovs: Replaced it with `Background` (not sure if its a great choice, other parts of the code just use `TODO`)

alvaroaleman (approved):
/assign @bradmwilliams

prow.go
bradmwilliams: Unfortunately, as @vrutkovs points out, there currently is not any relevant context configured anywhere in the code.  I would say it should probably remain as `context.TODO` and we can open a card to improve the context support of the chat-bot, as a whole, and not just for this one particular call.
@stevekuznetsov You placed the hold on this PR for the lack of a context, does this sound OK to you?

bradmwilliams: I have created the following [ticket](https://issues.redhat.com/browse/DPCR-82)

bradmwilliams (approved):
/lgtm

PR #189, opened 2021-07-28
Use correct kconfig flag in run script

hasbro17:
The flag `--build-cluster-kubeconfigs` has been deprecated and has no effect.
The correct flag to use in the run script is `--build-cluster-kubeconfigs-location`.

Prior to this change you would get the following error on running cluster-bot locally since the `--build-cluster-kubeconfigs-location` flag was unset and would default to `/var/build-cluster-kubeconfigs`:
```console
$ BOT_TOKEN=... ./hack/run.sh
2021/07/28 15:56:31 unable to load build cluster configurations: unable to access location "/var/build-cluster-kubeconfigs": open /var/build-cluster-kubeconfigs: no such file or directory
```

### DIFF
--- a/hack/run.sh
+++ b/hack/run.sh
@@ -4,8 +4,9 @@ set -euo pipefail
 
 if [[ -z $BOT_TOKEN ]]; then echo "BOT_TOKEN var must be set"; exit 1; fi
 
-tmp_kk=$(mktemp)
-trap 'rm -rf $tmp_kk' EXIT
+tmp_dir=$(mktemp -d)
+tmp_kk=$tmp_dir/build.kubeconfig
+trap 'rm -rf $tmp_dir' EXIT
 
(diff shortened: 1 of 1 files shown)


hasbro17:
We probably should remove the no-op flag `--build-cluster-kubeconfigs`.
https://github.com/openshift/ci-chat-bot/blob/697cf29410035a3baad03ca9cf478278317a3bd1/main.go#L67

That would break the cluster-bot manifest for app.ci though which is unnecessarily using the outdated flag, so might be good update that as well.
https://github.com/openshift/release/blob/9be226110ed679984f251e54fd3361527db2a2be/clusters/app.ci/ci-chat-bot/ci-chat-bot.yaml#L176

bradmwilliams (approved):
/lgtm

PR #193, opened 2021-08-17
Local execution enhancements

bradmwilliams:
A few more folks have inquired about making changes to the cluster-bot and how to test it.  I started looking into the current `hack/run.sh` and found a few things that needed to be addressed:

1. Proper interrupt handling
The script would launch the cluster-bot properly, but ^C wouldn't actually break out of the running program.  I added signal handling to the initialization of the cluster-bot.
2. The `BuildClusterClientConfigMap` is not being populated with the necessary values for the various build clusters
I added logic to copy the users kubeconfig over to cluster specific (`build01.kubeconfig`, `build02.kubeconfig`, `vsphere.kubeconfig`) kubeconfigs.  Although this doesn't necessarily give users the ability to interrogate the environment on these clusters (they will see `Unauthorizied` messages in the log), it does setup the configuration for when we get the appropriate RBAC necessary on these clusters to do so.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"io/ioutil"
@@ -9,6 +10,7 @@ import (
 	"log"
 	"net/url"
 	"os"
+	"os/signal"
 	"path"
 	"regexp"
 	"sync"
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -43,12 +45,28 @@ type options struct {
 }
 
 func main() {
-	if err := run(); err != nil {
+	cxt, cancel := context.WithCancel(context.Background())
```
smarterclayton, line 48: Prow for a while was installing its own handlers.  Is this overriding those or in parallel?

bradmwilliams: I believe this would be in parallel.  These particular changes stem from a conversation started over in #185 where we were called out for not passing a legitimate context into the prow client when updating a prowjob. I figured I would lay the ground work with this PR and then disseminate the context later.

bradmwilliams:
This PR is out of date and we've made tons of changes that make it obsolete now, so closing.

PR #205, opened 2021-09-13
Add ovirt supported platform

vrutkovs:
Launch task was added in https://github.com/openshift/release/pull/21784.

Tested locally and its coming up correctly, must-gather and deprovision steps are executed as expected on shutdown


bradmwilliams (approved):
/lgtm

PR #206, opened 2021-09-13
Adding support for DPTP-2431

bradmwilliams:
Per https://issues.redhat.com/browse/DPTP-2431, DPTP is changing the `tag_specification` declarations to a new format:
```
releases:
  initial:
    integration:
      name: NAME
      namespace: NS
  latest:
    integration:
      include_built_images: true
      name: NAME
      namespace: NS 
```
This PR updates updates the chat-bot's usage of the `tag_specification` to the new format.

PR #208, opened 2021-09-14
Don't declare tag_specification

vrutkovs:
This fixes "build org/repo#num" commands. ci-operator no longer accepts noth `tag_specification` and `releases`.

I was not able to test it fully however


vrutkovs:
Closed in favor of #206

PR #215, opened 2021-10-14
to get auth credentials

miyadav:
Added query in FAQ , please add if it looks ok .. I feel sometimes new users feel if there is a way to check if auth is generated or not.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -9,3 +9,8 @@
 
    1. Authenticate to one of the CI Registires and update your pull secret as defined [here](https://docs.ci.openshift.org/docs/how-tos/use-registries-in-build-farm/#how-do-i-log-in-to-pull-images-that-require-authentication) 
    2. Follow the instructions [here](https://docs.openshift.com/container-platform/4.7/support/remote_health_monitoring/opting-out-of-remote-health-reporting.html#images-update-global-pull-secret_opting-out-remote-health-reporting).
+
+
+3. **It's been more than 30mins I did not get auth credentials yet, what do I do?**
+   /msg auth
+   It will return message "your cluster is still getting created" or the kube-config file in case cluster is launched successfully


docs/FAQ.md
```diff
@@ -8,4 +8,9 @@
 2. **How can I update the "Global Cluster Pull Secret", on a cluster-bot created cluster, to pull from one of the CI Build Farms?**
 
    1. Authenticate to one of the CI Registires and update your pull secret as defined [here](https://docs.ci.openshift.org/docs/how-tos/use-registries-in-build-farm/#how-do-i-log-in-to-pull-images-that-require-authentication) 
-   2. Follow the instructions [here](https://docs.openshift.com/container-platform/4.7/support/remote_health_monitoring/opting-out-of-remote-health-reporting.html#images-update-global-pull-secret_opting-out-remote-health-reporting).
\ No newline at end of file
+   2. Follow the instructions [here](https://docs.openshift.com/container-platform/4.7/support/remote_health_monitoring/opting-out-of-remote-health-reporting.html#images-update-global-pull-secret_opting-out-remote-health-reporting).
+
+
+3. **It's been more than 30mins I did not get auth credentials yet, what do I do?**
+   /msg auth
```
bradmwilliams, line 15: You should put a newline in between the Question and the answer.  This will prevent the MD processor from combining the messages onto a single line.

I would also recommend changing "/msg auth" to just:
`auth`

bradmwilliams (changes requested):

docs/FAQ.md
```diff
@@ -8,4 +8,9 @@
...
+   It will return message "your cluster is still getting created" or the kube-config file in case cluster is launched successfully
```
bradmwilliams, line 16: You'll also want a newline here.

bradmwilliams (changes requested):

bradmwilliams:
/close
Picked up changes in #248

PR #228, opened 2021-11-03
prow,manager: add support for upgrade jobs

AlexNPavel:
This PR adds the ability to run upgrade jobs on the modern config type.
At the time of writing this commit message, the `e2e-upgrade-all` test
type is unable to be accurately run using the step registry as the step
registry's script hardcodes the parallel conformance suite for the
`upgrade-conformance` test type, while the exisiting `e2e-upgrade-all`
chatbot test uses the full conformance suite. In this PR, the chatbot
will fall back to templates for that test type until we can reconcile
this difference.

/cc @bradmwilliams

### DIFF
--- a/prow.go
+++ b/prow.go
@@ -134,6 +134,34 @@ var envsForTestType = map[string][]envVar{
 		name:  "TEST_TYPE",
 		value: "suite-conformance",
 	}},
+	"e2e-upgrade": {{
+		name:  "TEST_TYPE",
+		value: "upgrade",
+	}},
+	// Currently, the openshift-e2e-test `upgrade-conformance` test type hardcodes the parallel conformance suite to run after the upgrade test.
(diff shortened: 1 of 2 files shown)


prow.go
```diff
@@ -437,6 +465,20 @@ func (m *jobManager) newJob(job *Job) error {
 			}
 			// CLUSTER_DURATION unused by tests; remove to prevent ci-operator from complaining
 			delete(matchedTarget.MultiStageTestConfiguration.Environment, "CLUSTER_DURATION")
+		}
+		if job.Mode == JobTypeUpgrade {
+			if job.Mode == JobTypeUpgrade {
```
bradmwilliams, line 470: Why?

bradmwilliams (changes requested):

prow.go
AlexNPavel: That was a small mistake when I was reorganizing some stuff in that section. Removed the double ifs

PR #234, opened 2021-11-16
allow use of upi as a launch parameter

rvanderp3:
The intention of this PR is to allow use of `upi` as a launch parameter.  Initially, it is desired that this be consumable for vSphere launches(https://github.com/openshift/release/pull/23670)


bradmwilliams:
/hold
Waiting on https://github.com/openshift/release/pull/23670

bradmwilliams (approved):

PR #240, opened 2021-12-03
Fixing pointer reference issue with workflow config

bradmwilliams:


main.go
```diff
@@ -243,7 +243,7 @@ func manageWorkflowConfig(path string, workflows *WorkflowConfig) {
 		rawConfig, err := ioutil.ReadFile(path)
 		if err != nil {
 			klog.Errorf("Failed to load workflow config file at %s: %v", path, err)
-		} else if err := yaml.Unmarshal(rawConfig, config); err != nil {
+		} else if err := yaml.Unmarshal(rawConfig, &config); err != nil {
```
arjunrn, line 246: You also need to change `config` to a local value type and then pass its pointer to `Unmarshal`. Like in the example [here](https://pkg.go.dev/gopkg.in/yaml.v2#Unmarshal).

So something like this:

```go
var config WorkflowConfig
rawConfig, err := ioutil.ReadFile(path)
if err != nil {
	klog.Errorf("Failed to load workflow config file at %s: %v", path, err)
} else if err := yaml.Unmarshal(rawConfig, &config); err != nil {
	klog.Errorf("Failed to unmarshal workflow config: %v", err)
}
```

bradmwilliams: `config` is a pointer to allow for the `nil` check at line 251.  I have tested the code and it's working as expected.

arjunrn: Ok, it does work but the [docs](https://pkg.go.dev/gopkg.in/yaml.v2#Unmarshal) explicitly states:

> The out parameter must not be nil.

So would not rely on this working in the future. A safer option would be to test the `Workflows` field for `nil` in line 251.

bradmwilliams: Updated!

PR #251, opened 2022-02-18
WIP: Add support for Windows nodes

jrvaldes:
This PR adds the `windows-nodes` variant to support launching ovn-hybrid clusters with Windows worker nodes.

### DIFF
--- a/prow.go
+++ b/prow.go
@@ -60,7 +60,7 @@ var supportedUpgradeTests = []string{"e2e-upgrade", "e2e-upgrade-all", "e2e-upgr
 var supportedPlatforms = []string{"aws", "gcp", "azure", "vsphere", "metal", "hypershift", "ovirt", "openstack"}
 
 // supportedParameters are the allowed parameter keys that can be passed to jobs
-var supportedParameters = []string{"ovn", "ovn-hybrid", "proxy", "compact", "fips", "mirror", "shared-vpc", "large", "xlarge", "ipv6", "preserve-bootstrap", "test", "rt", "single-node", "cgroupsv2", "techpreview", "upi", "crun", "nfv", "kuryr"}
+var supportedParameters = []string{"ovn", "ovn-hybrid", "proxy", "compact", "fips", "mirror", "shared-vpc", "large", "xlarge", "ipv6", "preserve-bootstrap", "test", "rt", "single-node", "windows-nodes", "cgroupsv2", "techpreview", "upi", "crun", "nfv", "kuryr"}
 
 // multistageParameters is the mapping of supportedParameters that can be configured via multistage parameters to the correct environment variable format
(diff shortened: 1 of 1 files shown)


prow.go
```diff
@@ -60,7 +60,7 @@ var supportedUpgradeTests = []string{"e2e-upgrade", "e2e-upgrade-all", "e2e-upgr
 var supportedPlatforms = []string{"aws", "gcp", "azure", "vsphere", "metal", "hypershift", "ovirt", "openstack"}
 
 // supportedParameters are the allowed parameter keys that can be passed to jobs
-var supportedParameters = []string{"ovn", "ovn-hybrid", "proxy", "compact", "fips", "mirror", "shared-vpc", "large", "xlarge", "ipv6", "preserve-bootstrap", "test", "rt", "single-node", "cgroupsv2", "techpreview", "upi", "crun", "nfv", "kuryr"}
+var supportedParameters = []string{"ovn", "ovn-hybrid", "proxy", "compact", "fips", "mirror", "shared-vpc", "large", "xlarge", "ipv6", "preserve-bootstrap", "test", "rt", "single-node", "windows-nodes", "cgroupsv2", "techpreview", "upi", "crun", "nfv", "kuryr"}
```
aravindhp, line 63: Given this is going to bring up only a single Windows nodes, `windows-node` will be more appropriate. Looking at https://github.com/openshift/release/pull/26391/ I think this will only work on AWS. So what happen for `launch 4.9.9 vsphere,windows-node`?

aravindhp (commented):
Thanks for working on this @jrvaldes

prow.go
jrvaldes: Good questions. 

> windows-node will be more appropriate
That was my initial approach too, proposed  "windows-nodes"  because  `cluster-bot` variations support key-value pairs, hence plan to leverage that in the future to control the desired number of Windows nodes.

 For example, the following `cluster-bot` command:
```
launch 4.9.5 aws,windows-nodes=2
```

will installs WMCO and creates a Windows machineSet with 2 replicas. Initial implementation defaults to 1, if value is provided.




> So what happen for launch 4.9.9 vsphere,windows-node?

Defaults to a [legacy configuration](https://github.com/openshift/ci-chat-bot/blob/a62ff45def81ca6d22091377a318ff8976f1b147/manager.go#L1193) .

aravindhp: I think the `legacy configuration` has to do with using the older templates instead of the step-registry.

jrvaldes:
Closed to leverage the new ci-chat-bot feature [workflow-launch](https://github.com/openshift/release/blob/master/core-services/ci-chat-bot/workflows-config.yaml) in https://github.com/openshift/release/pull/28387.

PR #252, opened 2022-02-22
Add workflow-upgrade command

vrutkovs:
This command is similar to `workflow-launch`, but focuses on upgrade workflows. It requires two inputs (`from` and `to`) to be specified.

I don't have permissions to test this on live cluster, but the job config is similar to `test upgrade` steps


slack.go
```diff
@@ -411,6 +386,73 @@ func (b *Bot) Start(manager JobManager) error {
...
+
+			to, err := parseImageInput(request.StringParam("to_image_or_version_or_pr", ""))
+			if err != nil {
+				response.Reply(err.Error())
+				return
+			}
+			if len(from) == 0 {
+				response.Reply("you must specify the target release")
+				return
+			}
```
bradmwilliams, line 417: Copy and paste problem.  This should check `to`

bradmwilliams (changes requested):

slack.go
vrutkovs: Good catch, fixed

bradmwilliams (approved):

PR #253, opened 2022-02-23
Bumping to version 4.11 by default

slashpai:


bradmwilliams:
/hold
We do not bump the default version until the current version has officially gone GA (March 9th).

PR #258, opened 2022-03-03
Converting to Prow's GitHub Client

bradmwilliams:
Switching over to Prow's GitHub Client and converting the PR logic to always use the SHA value from the HEAD of the `master|main` branch.


main.go
```diff
@@ -46,6 +48,13 @@ type options struct {
 	WorkflowConfigPath              string
 }
 
+func (o *options) Validate() error {
+	if err := o.GitHubOptions.Validate(false); err != nil {
+		return err
+	}
+	return nil
+}
+
```
petr-muller, line 57: Technically this is `return o.GitHubOptions.Validate(false)` ;)

manager.go
```diff
@@ -926,41 +896,15 @@ func (m *jobManager) resolveAsPullRequest(spec string) (*prowapiv1.Refs, error)
...
 
-	var pr GitHubPullRequest
-	if err := json.NewDecoder(resp.Body).Decode(&pr); err != nil {
-		return nil, fmt.Errorf("unable to retrieve pull request info: %v", err)
-	}
 	if pr.Merged {
 		return nil, fmt.Errorf("pull request %s has already been merged to %s", spec, pr.Base.Ref)
 	}
-	if !pr.Mergeable {
+	if !*pr.Mergable {
```
petr-muller, line 907: I'm not sure if this is guaranteed to be non-nil, I'd better do if `pr.mergeable != nil && !*pr.Mergable`

petr-muller (commented):
yeah, they Prow GH client usage looks good, some comments inline

petr-muller:
You may even consider using ghproxy now; )

main.go
bradmwilliams: Done!

manager.go
bradmwilliams: Done!

bradmwilliams:
> You may even consider using ghproxy now; )

We are already setup to do so:
https://github.com/openshift/release/blob/1d7d7c0155403804fe00215bd2844012a8892047/clusters/app.ci/ci-chat-bot/ci-chat-bot.yaml#L200-L201

Our previous implementation only honored the last parameter that was specified! ;)

AlexNPavel (approved):

PR #267, opened 2022-04-26
Update to golang 1.18 and update test-infra dep

smg247:
`ci-tools` needs an updated `test-infra` dependency, and thus `ci-chat-bot` gets one too. Since `test-infra` recently went to golang `1.18` we also need to update it here.

See update PR for `release-controller`: https://github.com/openshift/release-controller/pull/392


smg247:
We need https://github.com/openshift/release/pull/28143 for the linter to pass

bradmwilliams (approved):

PR #278, opened 2022-08-05
enable cluster access to initiators

hoxhaeris:


kaovilai:
this is awesome

hoxhaeris:
/hold
2 new Cluster-Bot App OAuth scopes should be added before merging:
- users:profile:read
- users:read:email

manager.go
```diff
@@ -937,6 +940,14 @@ func (m *jobManager) resolveAsPullRequest(spec string) (*prowapiv1.Refs, error)
 
 func (m *jobManager) resolveToJob(req *JobRequest) (*Job, error) {
 	user := req.User
+	var userID string
+	if !strings.HasSuffix(req.UserProfile.Profile.Email, "@redhat.com") {
+		klog.Infof("failed to parse the UserID from the UserProfile email (not a redhat.com email)")
```
osherdp, line 945: should it be ``Debugf`` maybe? I'm not entirely familiar with the logging scheme for this component but it seems more appropriate to write external-dependent/non-critical logs with the debug level

prow.go
```diff
@@ -834,13 +836,61 @@ func getClusterClient(m *jobManager, job *Job) (*BuildClusterClientConfig, error
...
+		if job.RequesterUserID != "" {
+			client, err := ctrlruntimeclient.New(clusterClient.CoreConfig, ctrlruntimeclient.Options{})
+			if err != nil {
+				return false, err
+			}
+			if err := createAccessRBAC(namespace, job.RequesterUserID, client); err != nil {
+				klog.Infof("could not create role binding for %s: %v", job.RequesterUserID, err)
+				// the namespace might not yet exist when this step is executed
+				// we want to retry if this step fails, hence the nil return
+				return false, nil
```
osherdp, line 872: maybe it should be split to two conditions (for checking existence of the namespace and creation of rbac)?
that's because rbac creation might be because of other reasons

osherdp (commented):
looks great!
just added a couple of comments

prow.go
```diff
@@ -834,13 +836,61 @@ func getClusterClient(m *jobManager, job *Job) (*BuildClusterClientConfig, error
...
+		if err != nil {
+			return false, err
+		}
+		if job.RequesterUserID != "" {
+			client, err := ctrlruntimeclient.New(clusterClient.CoreConfig, ctrlruntimeclient.Options{})
+			if err != nil {
+				return false, err
+			}
+			if err := createAccessRBAC(namespace, job.RequesterUserID, client); err != nil {
+				klog.Infof("could not create role binding for %s: %v", job.RequesterUserID, err)
```
bradmwilliams, line 869: It's probably best if we dont start spraying folks RedHat ID's throughout our logs.  I'd recommend only printing their unique slack ID (`job.RequestedBy`).

prow.go
```diff
@@ -834,13 +836,61 @@ func getClusterClient(m *jobManager, job *Job) (*BuildClusterClientConfig, error
...
+				// the namespace might not yet exist when this step is executed
+				// we want to retry if this step fails, hence the nil return
+				return false, nil
+			}
+			klog.Infof("created the access RoleBinding for %s:", job.RequesterUserID)
```
bradmwilliams, line 874: It's probably best if we dont start spraying folks RedHat ID's throughout our logs. I'd recommend only printing their unique slack ID (`job.RequestedBy`).

prow.go
bradmwilliams: This particular error should probably be logged as an Errorf()

prow.go
```diff
@@ -834,13 +836,61 @@ func getClusterClient(m *jobManager, job *Job) (*BuildClusterClientConfig, error
...
+			return true, nil
+		}
+		return false, fmt.Errorf("failed to parse the RequesterUserID for %s", job.RequestedBy)
+	})
+	if err != nil {
+		klog.Infof("Failed to create the access role binding for %s: %v", job.RequestedBy, err)
```
bradmwilliams, line 880: This should probably be an Errorf() as well

prow.go
hoxhaeris: Fixed

prow.go
hoxhaeris: Fixed

prow.go
hoxhaeris: Fixed

manager.go
hoxhaeris: This was changed to `Errorf` instead.

prow.go
hoxhaeris: Generally, we want to retry on any type of failure.
The namespace one is noted since it occurs quite frequently (due to a race condition).

bradmwilliams (approved):

bradmwilliams (approved):

PR #291, opened 2022-08-25
OCPCRT-154: manager: add support for multiarch images

AlexNPavel:
This PR adds support for launching clusters using multiarch images from
https://multi.ocp.releases.ci.openshift.org. Clusters that use multiarch
images are launched on `amd64` based nodes and thus use the `amd64`
launch jobs.

### DIFF
--- a/manager.go
+++ b/manager.go
@@ -709,8 +709,6 @@ func (m *jobManager) resolveImageOrVersion(imageOrVersion, defaultImageOrVersion
 		// the release-controller cannot assemble multi-arch release, so we must use the `art-latest` streams instead of `release-multi`
 		imagestreams = append(imagestreams, namespaceAndStream{Namespace: "ocp-multi", Imagestream: "4.12-art-latest-multi", ArchSuffix: "-multi"})
 		imagestreams = append(imagestreams, namespaceAndStream{Namespace: "ocp-multi", Imagestream: "4.11-art-latest-multi", ArchSuffix: "-multi"})
-		imagestreams = append(imagestreams, namespaceAndStream{Namespace: "ocp-multi", Imagestream: "4.10-art-latest-multi", ArchSuffix: "-multi"})
-		imagestreams = append(imagestreams, namespaceAndStream{Namespace: "ocp-multi", Imagestream: "4.9-art-latest-multi", ArchSuffix: "-multi"})
 	default:
 		return "", "", "", fmt.Errorf("Unsupported architecture: %s", architecture)
 	}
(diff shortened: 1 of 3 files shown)


manager.go
```diff
@@ -692,36 +692,40 @@ func (m *jobManager) resolveImageOrVersion(imageOrVersion, defaultImageOrVersion
...
 	case "arm64":
-		imagestreams["ocp-arm64"] = "release-arm64"
-		archSuffix = "-arm64"
+		imagestreams = append(imagestreams, namespaceAndStream{Namespace: "ocp-arm64", Imagestream: "release-arm64", ArchSuffix: "-arm64"})
+	case "multi":
+		// the release-controller cannot assemble multi-arch release, so we must use the `art-latest` streams instead of `release-multi`
+		imagestreams = append(imagestreams, namespaceAndStream{Namespace: "ocp-multi", Imagestream: "4.12-art-latest-multi", ArchSuffix: "-multi"})
+		imagestreams = append(imagestreams, namespaceAndStream{Namespace: "ocp-multi", Imagestream: "4.11-art-latest-multi", ArchSuffix: "-multi"})
+		imagestreams = append(imagestreams, namespaceAndStream{Namespace: "ocp-multi", Imagestream: "4.10-art-latest-multi", ArchSuffix: "-multi"})
+		imagestreams = append(imagestreams, namespaceAndStream{Namespace: "ocp-multi", Imagestream: "4.9-art-latest-multi", ArchSuffix: "-multi"})
```
bradmwilliams, line 713: Is this something we're going to have to update every release?

AlexNPavel: Yes. We will need to add the new nightly imagestreams when the become available since there isn't a single imagestream that tags them all, unlike the amd64 and arm64 release streams.

bradmwilliams: Ack!  We might need to think about only adding the new entry when we bump the default version up (i.e. once we GA the current release) or we'll potentially be using an N and an N+1 version to launch.

bradmwilliams: We might want to update the `lookup` command's output to handle the scenarios where more than 1 images is necessary to launch a cluster.

deepsm007: We do not need 4.9/4.10 as it is not officially supported

AlexNPavel: Removed the 4.9 and 4.10 imagestreams from multiarch

bradmwilliams:
/lgtm
/hold
Waiting until someone from our team is available in case there are any issues.  Please `/unhold` whenever you're ready.

PR #304, opened 2022-10-05
add a retry mechanism for PostMessage

hoxhaeris:
On certain occasions (very rarely), posting a response to the requestor fails (non-client-related reasons). 
This PR adds a retry for posting the response to the requestor.

### DIFF
--- a/eventHandler.go
+++ b/eventHandler.go
@@ -5,9 +5,11 @@ import (
 	"github.com/slack-go/slack"
 	"github.com/slack-go/slack/slackevents"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/pkg/version"
 	"k8s.io/klog"
 	"strings"
+	"time"
 )
 
 func HandleEvent(client *slack.Client, callback *slackevents.EventsAPIEvent, manager JobManager, botCommands []BotCommand) error {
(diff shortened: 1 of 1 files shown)


eventHandler.go
```diff
@@ -43,19 +45,32 @@ func HandleEvent(client *slack.Client, callback *slackevents.EventsAPIEvent, man
...
+	}
+	return nil
+}
+
+func postResponse(client *slack.Client, event *slackevents.MessageEvent, response string) error {
+	var lastErr error
+	err := wait.PollImmediate(5*time.Second, 20*time.Second, func() (bool, error) {
+		_, responseTimestamp, err := client.PostMessage(event.Channel, slack.MsgOptionText(response, false))
+		if err != nil {
+			klog.Errorf("Failed to post response to UserID: %s; (event: `%s`) at %d; %v", event.User, event.Text, (time.Now()).Unix(), err)
```
bradmwilliams, line 65: This seems like it'll be pretty chatty if/when this happens.  Does the `event.Text` provide anything of value when this call fails?  or can this log message be made at line 73 instead?

bradmwilliams (changes requested):

eventHandler.go
hoxhaeris: Moved to line 73. Only the `err` variable does provide something of value (in theory there might be a different error on each try, but that is highly unlikely). The `event.Text` contains the given command text only

bradmwilliams (approved):

PR #307, opened 2022-10-05
slack: Make it easier to self-help workflow-launch failures

wking:
By listing known workloads and pointing folks at where they can (hopefully) add more.  Assuming the config comes from the release repo is brittle, but since we're probably the only folks running this thing, it may help.  I guess we could also add an explicit config knob for "where to suggest changes to this config", but I'm punting on that for now.

### DIFF
--- a/slack.go
+++ b/slack.go
@@ -8,6 +8,7 @@ import (
 	"k8s.io/klog"
 	"math/rand"
 	"net/http"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


wking:
> I guess we could also add an explicit config knob for "where to suggest changes to this config", but I'm punting on that for now.

[Here][1] is the release config setting `--workflow-config-path=/etc/workflows-config/workflows-config.yaml`, but working from there to the GitHub URI would be fiddly.  Still, `--workflow-config-path` would be a pattern to follow, if we wanted a `--workflow-config-reference=${URI}` or some such.

hoxhaeris (approved):

AlexNPavel (approved):

PR #312, opened 2022-10-26
add router&handler for events + refactoring

hoxhaeris:


main.go
```diff
@@ -150,7 +142,7 @@ func run() error {
 		return fmt.Errorf("the environment variable BOT_SIGNING_SECRET must be set")
 	}
 
-	prowJobKubeconfig, _, _, err := loadKubeconfig()
+	prowJobKubeconfig, _, _, err := manager.LoadKubeconfig()
```
bradmwilliams, line 145: While I understand how these changes came about, I dont think the methods related to loading of kubeconfigs should be part of the `manager` package.  They are really only used in `main` and really just muddy up the manager code.

main.go
```diff
@@ -161,7 +153,7 @@ func run() error {
 	prowClient := dynamicClient.Resource(schema.GroupVersionResource{Group: "prow.k8s.io", Version: "v1", Resource: "prowjobs"})
 
 	// Config and Client to access release images
-	releaseConfig, err := loadKubeconfigFromFlagOrDefault(opt.ReleaseClusterKubeconfig, prowJobKubeconfig)
+	releaseConfig, err := manager.LoadKubeconfigFromFlagOrDefault(opt.ReleaseClusterKubeconfig, prowJobKubeconfig)
```
bradmwilliams, line 156: Same as above.  Loading kubeconfigs shouldnt be part of the Manager.

pkg/manager/manager.go
```diff
@@ -242,7 +262,7 @@ func (m *jobManager) Start() error {
 	return nil
 }
 
-func paramsFromAnnotation(value string) (map[string]string, error) {
+func ParamsFromAnnotation(value string) (map[string]string, error) {
```
bradmwilliams, line 265: This method feels like it belongs in a more common/shared place since it's used in multiple packages, but really has nothing to do, specifically, with the `manager`

pkg/manager/manager.go
```diff
@@ -488,7 +508,7 @@ func (m *jobManager) estimateCompletion(requestedAt time.Time) time.Duration {
 	return lastEstimate.Truncate(time.Second)
 }
 
-func contains(arr []string, s string) bool {
+func Contains(arr []string, s string) bool {
```
bradmwilliams, line 511: This method feels like it belongs in a more common/shared place since it's used in multiple packages, but really has nothing to do, specifically, with the `manager`

pkg/manager/prow.go
```diff
@@ -1327,11 +1328,11 @@ func commandContents(podClient coreclientset.CoreV1Interface, podRESTConfig *res
 	return buf.String(), nil
 }
 
-// loadKubeconfig loads connection configuration
+// LoadKubeconfig loads connection configuration
 // for the cluster we're deploying to. We prefer to
 // use in-cluster configuration if possible, but will
 // fall back to using default rules otherwise.
-func loadKubeconfig() (*rest.Config, string, bool, error) {
+func LoadKubeconfig() (*rest.Config, string, bool, error) {
```
bradmwilliams, line 1335: See comments in `main.go`

pkg/manager/manager.go
```diff
@@ -161,6 +163,24 @@ func (j Job) IsComplete() bool {
...
+	Platform     string                                     `yaml:"platform"`
+}
+
+// LeaseClient only include the metrics function, as we don't want to create leases
+type LeaseClient interface {
+	// Metrics queries the states of a particular resource, for informational
+	// purposes.
+	Metrics(rtype string) (lease.Metrics, error)
+}
+
```
bradmwilliams, line 183: Since we are refactoring into packages, it's probably best to introduce a `pkg/manager/types.go` instead of carrying everything around hidden in all the nooks and crannies.

bradmwilliams (changes requested):

main.go
hoxhaeris: Moved the method to `main.go`

main.go
hoxhaeris: Moved the method to `main.go`

pkg/manager/manager.go
hoxhaeris: Moved the function to `utils.go`.

pkg/manager/manager.go
hoxhaeris: Moved the function to `utils.go`

pkg/manager/prow.go
hoxhaeris: Moved to `main.go`.

pkg/manager/manager.go
hoxhaeris: All types are moved to `types.go`.

hoxhaeris:
/cc @AlexNPavel 
Not sure why Alex was removed from the reviews list.

pkg/manager/prow.go
```diff
@@ -7,6 +7,8 @@ import (
 	"encoding/base32"
 	"encoding/json"
 	"fmt"
+	utils "github.com/openshift/ci-chat-bot/pkg"
```
AlexNPavel, line 10: Should be named `util` or be unnamed for consistency with the other files in the project.

AlexNPavel (changes requested):
Overall it looks good. However, I would suggest that the `utils.go` file be under its own subdirectory (i.e. `/pkg/utils`) instead of being top-level in `pkg` to make the imports clearer.

pkg/manager/manager.go
```diff
@@ -1,18 +1,19 @@
-package main
+package manager
 
 import (
 	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
+	utils "github.com/openshift/ci-chat-bot/pkg"
```
AlexNPavel, line 9: Should be named `util` or be unnamed for consistency with the other files in the project.

main.go
```diff
@@ -3,12 +3,14 @@ package main
 import (
 	"flag"
 	"fmt"
+	"github.com/openshift/ci-chat-bot/pkg"
```
AlexNPavel, line 6: Other files have this renamed to `utils`, but this import using the package name of `util`. Either this should also be named `utils` or the other imports should use the package name. I think using the package name would be preferable.

hoxhaeris: Done

hoxhaeris:
/hold
The slack methods should be moved to `pkg/slack.go`, along with the command/action functions, since we need to reuse those when introducing interactive workflows.

hoxhaeris:
A few more changes that came with the last commit ([e0a174f](https://github.com/openshift/ci-chat-bot/pull/312/commits/e0a174f657bc5f1d1c29f1cc35e7f5e5bb9e9b85)):
- Move the `Bot` type and methods to  `pkg/slack`, along with the functions that define the actions (Launch, List, etc)
- Move LoadKubeconfig and LoadKubeconfigFromFlagOrDefault from `main.go` to `pkg/utils/utils.go` (reusable code, keeping the main tidy)
- Clean up `pkg/prow/prow.go`
-  Move `commandParser.go` to a different subdirectory, to avoid circular importing
- Separate actions from `slack.go`, to a separate file (`actions.go`)

hoxhaeris:
> The slack methods should be moved to `pkg/slack.go`, along with the command/action functions, since we need to reuse those when introducing interactive workflows.

This is completed (I will keep the `hold`, pending review).

AlexNPavel (approved):

PR #319, opened 2022-11-15
Remove support for template jobs

AlexNPavel:
This PR removes all support for template files and performs some cleanup based on recommendations from golangci-lint.


bradmwilliams:
/lgtm
/hold
@AlexNPavel This is great! Feel free to `/unhold` when you return.

PR #350, opened 2023-02-03
update interactive `launch`

hoxhaeris:
Add a few in-between steps when selecting the version, to avoid having to redirect users to the `release-controller` page.

### DIFF
--- a/pkg/slack/modals/launch/launch.go
+++ b/pkg/slack/modals/launch/launch.go
@@ -17,7 +17,7 @@ import (
 
 const (
 	IdentifierInitialView        modals.Identifier = "launch"
-	Identifier3rdStep            modals.Identifier = "launch3ddStep"
+	Identifier3rdStep            modals.Identifier = "launch3rdStep"
 	IdentifierPRInputView        modals.Identifier = "pr_input_view"
 	IdentifierFilterVersionView  modals.Identifier = "filter_version_view"
 	IdentifierRegisterLaunchMode modals.Identifier = "launch_mode_view"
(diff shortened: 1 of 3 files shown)


pkg/slack/modals/launch/launch.go
```diff
@@ -8,15 +8,44 @@ import (
...
 	"sync"
 )
 
-// Identifier is the view identifier for this modal
-const Identifier modals.Identifier = "launch"
-const Identifier2ndStep modals.Identifier = "launch2ndStep"
-const Identifier3rdStep modals.Identifier = "launch3ddStep"
+const (
+	IdentifierInitialView        modals.Identifier = "launch"
+	Identifier3rdStep            modals.Identifier = "launch3ddStep"
```
bradmwilliams, line 20: Should `launch3ddStep` be `launch3rdStep`?

pkg/slack/modals/launch/launch.go
```diff
@@ -8,15 +8,44 @@ import (
...
+	launchArchitecture   = "architecture"
+	launchParameters     = "parameters"
+	launchVersion        = "version"
+	launchStepContext    = "context"
+	defaultPlatform      = "aws"
+	defaultArchitecture  = "amd64"
+	launchMode           = "launch_mode"
+	launchModeVersion    = "version"
+	launchModePR         = "pr"
+	launchModePRKey      = "One or multiple PR-s"
```
bradmwilliams, line 45: `PR-s` seems weird to me.  Can it not `PR's` or just `PR's`?

bradmwilliams (changes requested):

pkg/slack/modals/launch/launch.go
hoxhaeris: Fixed

pkg/slack/modals/launch/launch.go
hoxhaeris: Fixed

bradmwilliams (approved):

PR #361, opened 2023-04-07
Enable the cluster profile to be specified for workflow-launch

rvanderp3:
For some platforms, particularly vSphere, there are numerous environment variations that require that specific workflows run in their associated lease pools.  The intent of this PR is to introduce a parameter to workflow-launch to allow a specific cluster profile to chosen.

Alternatively, additional launch variants could be created for vSphere but this would add numerous variants which are entirely irrelevant to other platforms.  Additionally, this makes the help more confusing and requires additional maintenance to add a new variant each time we introduce a new  configuration or environment to test.

### DIFF
--- a/pkg/manager/prow.go
+++ b/pkg/manager/prow.go
@@ -439,14 +439,18 @@ func (m *jobManager) newJob(job *Job) (string, error) {
 	if job.Mode == JobTypeWorkflowLaunch || job.Mode == JobTypeWorkflowUpgrade {
 		// use "launch" test name to identify proper cluster profile
 		var profile citools.ClusterProfile
-		for _, test := range sourceConfig.Tests {
-			if test.As == "launch" {
-				profile = test.MultiStageTestConfiguration.ClusterProfile
-			}
-		}
 		environment := citools.TestEnvironment{}
 		for name, value := range job.JobParams {
 			environment[name] = value
+		}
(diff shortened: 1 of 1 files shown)


bradmwilliams:
/hold
Unfortunately, while the theory behind this logic is sound, it does not account for the lease balancing logic, via Boskos, that we rely on further down in the code.  We'll talk this over, as a team, in our scrum and update the PR accordingly.

rvanderp3:
> /hold Unfortunately, while the theory behind this logic is sound, it does not account for the lease balancing logic, via Boskos, that we rely on further down in the code. We'll talk this over, as a team, in our scrum and update the PR accordingly.

Thanks for looking over it an considering it @bradmwilliams

rbbratta:
@bradmwilliams this could be useful to unblock vsphere dualstack testing which requires the specific `vsphere-ibm-7`.  Have you had a chance to look at it?

bradmwilliams:
My apologies for letting this PR fall off the radar.  We talked this over, as a team, and everyone was in agreement that allowing users to override `CLUSTER_PROFILE` introduces a lot of risk as it is a core variable to the CI environment.
I know that the variant approach might be a bit more work (2 PRs instead of 1), but it lends itself to a more consistent approach with all the other platforms and doesn't expose CLUSTER_PROFILE to the world.

PR #362, opened 2023-04-13
refactor Interactive `launch`

hoxhaeris:
Refactor some of the steps on the interactive `launch`.
Improve the error feedback and input validation.


pkg/slack/modals/launch/steps/prInput.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	if len(prErrors) == 0 {
+		return nil
+	}
+
+	errors[launch.LaunchFromPR] = strings.Join(prErrors, "; ")
+	response, err := modals.ValidationError(errors)
+	if err != nil {
+		klog.Warningf("failed to build validation error: %v", err)
```
bradmwilliams, line 99: Have you actually seen this log message come through in a log?  I've seen issues where jumping between `k8s.io/klog` and `k8s.io/klog/v2` causes the V2 messages to not be printed into the log.

hoxhaeris: Looks like it is logging. Nonetheless, I will change it to `k8s.io/klog`, to be consistent with the rest.

bradmwilliams (approved):

PR #391, opened 2023-05-29
update vendors, use the test-infra fork

droslean:
Signed-off-by: Nikolaos Moraitis <nmoraiti@redhat.com>


droslean:
/close

Temporarily it would be better stop synchronizing the test-infra package with ci-tools, until we start using the upstream test-infra instead of the fork.

PR #396, opened 2023-06-08
manager: support indexless bundle configurations

AlexNPavel:
This PR adds support for launching clusters with optional operators that do not build an index and instead use a step that manually runs `operator-sdk run bundle`. Steps will have to match the general naming scheme specified on docs.ci.openshift.org. After this PR is merged, the docs website will be updated to include this information.


bradmwilliams (approved):

AlexNPavel:
Fixes OCPCRT-218

PR #416, opened 2023-09-20
Bumping to the latest/greatest test-infra

bradmwilliams:
I made an update to `test-infra` for a new Jira API.  This PR is to pick that up.  It might fail CI though...
I can't remember if we're allowed to be ahead of `ci-tools` or not.  So, we're going to find out.


bradmwilliams:
/hold
Just in case

bradmwilliams:
/unhold
Verified with DPTP that we are good to take a `test-infra` bump without also bumping `ci-tools`.

AlexNPavel (approved):

PR #427, opened 2023-10-23
(re)Enable rosa integration

bradmwilliams:
Reverting my previous changes.


bradmwilliams:
/hold
The incident has been declared resolved, but we have a handful of clusters that need to be cleaned up before we turn everything back on.

hoxhaeris (approved):

PR #438, opened 2023-11-23
Indicate 'build' command can use base image

stephenfin:
...and indicate default image version used for 'launch'


stephenfin:
Fixed unit test issue. `ci/prow/security` failure is due to us hitting snyk API limits

PR #444, opened 2023-12-20
ci-operator: promote multiple targets

danilo-gemoli:
Test Platform team is introducing a new feature in `ci-operator`: multi-target promotion https://github.com/openshift/ci-tools/pull/3858.
This PR if a follow up of the aforementioned work in order to keep the `ci-operator` configuration scheme up to date in `ci-chat-bot` too.

As of today the `ci-operator` promotion stanza looks like:
```yaml
promotion:
  namespace: ci
  tag: latest
```
With the new feature we are moving toward the following schema:
```yaml
promotion:
  to:
  - namespace: ci
    tag: latest
  - namespace: ocp
    tag: 4.16
  ...
```
that allows promoting to more than a single place.

Both schema are supported but we count to deprecate the former somewhere, not too far away, in the future.

This PR is huge only because I had to update the module `github.com/openshift/ci-tools` to the `latest` version (first two commits), otherwise the scope would have limited to a single file: `pkg/manager/prow.go`.


hoxhaeris (approved):

hoxhaeris:
/hold
@danilo-gemoli, feel free to unhold when ready.

danilo-gemoli:
Snyk is complaning about [this username](https://github.com/openshift/ci-chat-bot/blob/master/pkg/manager/rosa.go#L357C17-L357C17) being [hardcoded](https://github.com/openshift/ci-chat-bot/blob/deac73015e4c8eea025441c74272b22f34f8a560/pkg/manager/rosa.go#L36) and therefore publicly exposed.
At a first glance it doesn't look like a big deal but I'm not sure whether ignoring the warning and moving forward anyway or not.

bradmwilliams (approved):

bradmwilliams:
> Snyk is complaning about [this username](https://github.com/openshift/ci-chat-bot/blob/master/pkg/manager/rosa.go#L357C17-L357C17) being [hardcoded](https://github.com/openshift/ci-chat-bot/blob/deac73015e4c8eea025441c74272b22f34f8a560/pkg/manager/rosa.go#L36) and therefore publicly exposed. At a first glance it doesn't look like a big deal but I'm not sure whether ignoring the warning and moving forward anyway or not.

This shouldn't be an issue.  `cluster-admin` is simply the name we use when generating the login credentials for the cluster being created.  The password is what's really important.  It's randomly generated and never exposed publicly.

PR #447, opened 2024-02-12
Add "workflow-test" command

vrutkovs:
This adds new "workflow-test" command, which is similar to `workflow-launch`, but doesn't mutate tests.

This is useful to run a particular workflow for a set of pull-requests or with varied parameters

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -49,8 +49,17 @@
 
    The workflow config file also supports setting base_images for workflows that may require special base images to run as well as an architecture field that will support configuring the CPU architecture for openshift to run on once non-amd64 architectures are supported by the ci-chat-bot (https://github.com/openshift/ci-chat-bot/blob/master/prow.go#L190).
 
+## workflow_test
+1. **What is `workflow_test`?**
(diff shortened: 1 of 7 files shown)


docs/FAQ.md
```diff
@@ -49,8 +49,17 @@
 
    The workflow config file also supports setting base_images for workflows that may require special base images to run as well as an architecture field that will support configuring the CPU architecture for openshift to run on once non-amd64 architectures are supported by the ci-chat-bot (https://github.com/openshift/ci-chat-bot/blob/master/prow.go#L190).
 
+## workflow_test
+1. **What is `workflow_test`?**
 
-2. **How can I make a workflow runnable by the `workflow-launch` command**
+   Lorem ipsu,
```
bradmwilliams, line 55: Appreciate the reference, but probably not very helpful for answering the question. ;)

bradmwilliams (changes requested):
Changes look good!  Could you just update the `lorem ipsum`?

bradmwilliams (approved):

PR #455, opened 2024-03-14
update `test-infra` dependency

danilo-gemoli:
Update `go get k8s.io/test-infra@v0.0.0-20240314024253-449bac00dc77` as we need to bump it in `o/ci-tools` too, versions have to match.


danilo-gemoli:
Needed for https://github.com/openshift/ci-tools/pull/4047

PR #457, opened 2024-03-20
NO-JIRA: Introduce new oc get execution command

ardaguclu:
This PR introduces new command to execute readonly oc commands in Slack against launched clusters by using the generated sole kubeconfig. Currently, only `oc get` will be allowed to collect
some feedback and in the future, this can be extended.

I'd like to note that this is just a PoC and haven't been tested yet.

### DIFF
--- a/pkg/slack/actions.go
+++ b/pkg/slack/actions.go
@@ -418,6 +418,21 @@ func WorkflowUpgrade(client *slack.Client, jobManager manager.JobManager, event 
 	return msg
 }
 
+func ExecuteOC(client *slack.Client, jobManager manager.JobManager, event *slackevents.MessageEvent, properties *parser.Properties) string {
+	cluster, password := jobManager.GetROSACluster(event.User)
+	if cluster != nil {
+		ExecuteOCOnRosa(client, cluster, password, event.Text)
+		return ""
+	}
+	job, err := jobManager.GetLaunchJob(event.User)
+	if err != nil {
+		return err.Error()
+	}
+	job.RequestedChannel = event.Channel
+	ExecuteOCOnJob(client, job, event.Text)
+	return ""
+}
+
 func RosaCreate(client *slack.Client, jobManager manager.JobManager, event *slackevents.MessageEvent, properties *parser.Properties) string {
 	from, err := ParseImageInput(properties.StringParam("version", ""))
 	if err != nil {
--- a/pkg/slack/slack.go
+++ b/pkg/slack/slack.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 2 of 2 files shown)


ardaguclu:
@bradmwilliams @AlexNPavel I spent some time in shiftweek to this feature and I'd like to hear your opinions about this, when you have a chance?. Would it make sense (or have you considered this feature already?) or unnecessarily increase complexity?. Thanks.

bradmwilliams:
Hey @ardaguclu,

> Would it make sense (or have you considered this feature already?) 

We have not had any requests for such functionality and this really isn't anything we've discussed previously.  I'm certainly curious how/why you've gone down this route.  Is there some situation you've encountered or a scenario that you think these types of commands would solve?  Or was this a `it would be really cool if the clusterbot did this` type decision?

> unnecessarily increase complexity?

My initial thought here is: probably so...
We currently allow up-to 80 traditional clusters and 15 Rosa clusters (hopefully more very soon) at any one time.  The complexity will come in efficiently/effectively handling the lifecycle of all the various kubeconfigs in a way that allows for us to make subsequent calls (because someone who uses this style of command probably isn't just going to use it once).  Which leads to the next issue of the supported commands.  As we all know `oc` has an enormous amount of command and sub-command options.  It will be nearly impossible for us to support everything that everyone might want to do with this type of command exposed via Clusterbot.

The main concerns I have are:
1. Clusterbot interacting with clusters, that it created, has never been something we've done to this point (i.e. we don't even have `oc` in our container).  That's certainly not a deal-breaker, but there is an inherent burden in keeping it up to date.  
2. How well does it support all the various cluster/variant combinations (i.e. will a pod running on the `app.ci` cluster be able to communicate with every cluster type we support via an `oc --kubeconfig` call).
3. Enabling something like this will inevitably result in us having to answer the "Why don't you support `oc foo`?" or "Why is `oc get super-fancy-awesomeness` not supported?", etc.  We could rely on the open-source model to allow folks to add the necessary support for whatever command their interested in.  However, we don't have any way for them to test that it actually works. 

Ultimately, I think the idea is cool and would provide a nice chat-bot style interface to an Openshift cluster, but I'm not sure that we'll be able to give it the time/support it needs to really get it off the ground without significant effort. 

That being said, this is only my opinion.  I'm certainly interested in what the others on the team have to say...
@openshift/openshift-team-continuous-release-team

ardaguclu:
Thanks for dropping your opinions and it all makes sense in terms of usability, maintainability, etc. That PR was indeed `it would be really cool if the clusterbot did this type decision` and it can serve as a historical discussion point in case there will be such discussions in the future.

I'm closing this, it seems that implementing standalone bot just this particular purpose is better;
/close

PR #469, opened 2024-04-17
slack: default arch to "multi" for hypershift-hosted

AlexNPavel:
See: https://docs.ci.openshift.org/docs/architecture/ci-operator/#testing-with-a-cluster-from-hypershift

Requires https://github.com/openshift/release/pull/51069

### DIFF
--- a/pkg/slack/slack.go
+++ b/pkg/slack/slack.go
@@ -448,12 +448,9 @@ func ParseOptions(options string, inputs [][]string, jobType manager.JobType) (s
 			return "", "", nil, fmt.Errorf("unrecognized option: %s", opt)
 		}
 	}
-	if len(architecture) == 0 {
-		architecture = "amd64"
-	}
 	if len(platform) == 0 {
 		switch architecture {
-		case "amd64":
+		case "", "multi":
 			// for hypershift, only support normal launches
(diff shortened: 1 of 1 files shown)


pkg/manager/prow.go
```diff
@@ -413,6 +413,22 @@ func (m *jobManager) newJob(job *Job) (string, error) {
...
+				case "RELEASE_IMAGE_LATEST":
+					exists = true
+					c.Env[j].Value = image
+				}
+			}
+			if !exists {
+				c.Env = append(c.Env, corev1.EnvVar{Name: "RELEASE_IMAGE_LATEST", Value: image})
+			}
+		}
+	}
```
bradmwilliams, line 431: Is this logic just to ensure that `RELEASE_IMAGE_LATEST` gets set in the event that the environment wasn't updated in the `prow.OverrideJobEnvironment(...)` call above?

bradmwilliams: Line 397

AlexNPavel: I was just copying what we did for arm64, but it was unnecessary since we are using the normal release image env var for hypershift. I've removed this bit of code. I tested the change and everything still works as expected.

bradmwilliams (approved):

PR #481, opened 2024-07-08
Allowing `workflow-launch` to create multi-arch `hypershift-hosted` clusters

bradmwilliams:
Folks have been struggling to launch `hypershift-hosted` clusters via `workflow-launch` because the way we select the underlying job that will house the injected workflow configuration.  This fix will allow us to pass the `hypershift-hosted` platform directly through to the job selector and we should pick up the correctly defined job and then inject the specified workflow.

This PR is only the first part of the fix.  Once merged, I will chase it with a change to the workflow-config.yaml, in openshift/release, to update the `hypershift-hostedcluster-workflow`'s  platform to: `hypershift-hosted`.


AlexNPavel (approved):

bradmwilliams:
Second PR to enable these changes: https://github.com/openshift/release/pull/54124

PR #508, opened 2024-09-25
*: add support for MCE clusters

AlexNPavel:
This PR adds support for authorized users to create clusters using MCE on the DPCR cluster.


pkg/manager/mce.go
```diff
@@ -0,0 +1,373 @@
...
+			AWS: &aws.MachinePool{
+				Zones: []string{"us-east-1a"},
+				EC2RootVolume: aws.EC2RootVolume{
+					IOPS: 2000,
+					Size: 100,
+					Type: "io1",
+				},
+				InstanceType: "m5.xlarge",
+			},
+		}
```
bradmwilliams, line 147: We should probably bump up to `m6a` or even `m6g` (if possible?), because they are more modern and cheaper.

pkg/manager/mce.go
```diff
@@ -0,0 +1,373 @@
...
+	kubeconfigSecret, err := secretsClient.Get(context.TODO(), clusterDeployment.Spec.ClusterMetadata.AdminKubeconfigSecretRef.Name, metav1.GetOptions{})
+	if err != nil {
+		return "", "", err
+	}
+	/*
+		kubeconfig, err := base64.RawStdEncoding.DecodeString(string(kubeconfigSecret.Data["kubeconfig"]))
+		if err != nil {
+			return "", "", fmt.Errorf("failed to decode kubeconfig: %v", err)
+		}
+	*/
```
bradmwilliams, line 341: Just want to make sure we don't need the decoded strings here?  Can probably be deleted if not.

pkg/manager/mce.go
```diff
@@ -0,0 +1,373 @@
...
+	passwordSecret, err := secretsClient.Get(context.TODO(), clusterDeployment.Spec.ClusterMetadata.AdminPasswordSecretRef.Name, metav1.GetOptions{})
+	if err != nil {
+		return "", "", err
+	}
+	/*
+		password, err := base64.RawStdEncoding.DecodeString(string(passwordSecret.Data["password"]))
+		if err != nil {
+			return "", "", fmt.Errorf("failed to decode password: %v", err)
+		}
+	*/
```
bradmwilliams, line 351: Same

bradmwilliams (changes requested):

pkg/manager/mce.go
AlexNPavel: We don't need to decode the strings. I've removed this code

pkg/manager/mce.go
AlexNPavel: Changed to `m6a`.
`m6g` is arm64 based, so we might not want to use that as the default. Maybe it can be an option later on

bradmwilliams: Yep.  The arm64 nodes are just slightly cheaper, that's all.  Having them as an option is a good idea.  Do you know if we might be able to utilize spot instances for worker nodes?

AlexNPavel:
The PR has been rebased for the go.mod changes from the rosa bump

bradmwilliams:
> The PR has been rebased for the go.mod changes from the rosa bump

Awesome!  Thanks!

pkg/manager/mce.go
AlexNPavel: It doesn't look like you can configure the cluster to use spot instances via the install-config.yaml. Our prowjobs do it by outputting the yaml of all manifests, modifying them with yq, and then passing the updated manifests to openshift-install. There's no way in MCE's web UI either.

bradmwilliams: Yeah, I was afraid of that.  Oh well.  Hopefully, it's something MCE is working on.

PR #517, opened 2024-10-16
manager: use prow cluster scheduler

AlexNPavel:


bradmwilliams (approved):
/hold
Let's let the MCE PR merge and roll out cleanly first... I want to make sure we have eyes on this PR when it deploys... just in case it falls over.

PR #520, opened 2024-10-21
manager: keep prowjob scheduler in order to use cache

AlexNPavel:
Store a scheduler interface as part of the manager struct. This allows the scheduler to use its built in cache, whose timeout is defined in the prow config (currently 10 minutes). The chat-bot likely won't make too much use of it, but it can't hurt to potentially reduce the load on the scheduler.


bradmwilliams (approved):
/hold
Until we have someone who can watch it roll out

PR #535, opened 2025-01-24
OCPCRT-360:  pass values to DEVSCRIPTS_CONFIG_CLUSTERBOT to support overrides

DennisPeriquet:
The `DEVSCRIPTS_CONFIG` variable is used to pass an array of `key=value` pairs to [some code](https://github.com/openshift/release/blob/7c9e17caab51f23303d648a5f2f491f770c72c35/ci-operator/step-registry/baremetalds/devscripts/setup/baremetalds-devscripts-setup-commands.sh#L171-L178) that extracts those `key=value` pairs as environment variables.

Since we cannot pass an array from a clusterbot command line to the prow job that creates the Openshift cluster, we pass a separate space-separated list of `key=value` pairs.

This PR introduces that space-separated list in a variable called `DEVSCRIPTS_CONFIG_CLUSTERBOT`. For example, in the clusterbot app, you can say: `workflow-launch cucushift-installer-rehearse-baremetalds-ipi-ovn 4.19 "DEVSCRIPTS_CONFIG_CLUSTERBOT=INSTALLER_PROXY=true\nIP_STACK=v4\nNETWORK_TYPE=OVNKubernetes`.

Because we introduce a new variable, we need this [release PR](https://github.com/openshift/release/pull/60930) to merged first.

### DIFF
--- a/pkg/slack/slack.go
+++ b/pkg/slack/slack.go
@@ -310,6 +310,25 @@ func BuildJobParams(params string) (map[string]string, error) {
 	}
 	jobParams := make(map[string]string)
 	for _, combinedParam := range splitParams {
+		// Check for DEVSCRIPTS_CONFIG_CLUSTERBOT ; if found, we know we're dealing with newline separated parameters.
+		if strings.Contains(combinedParam, "DEVSCRIPTS_CONFIG_CLUSTERBOT") {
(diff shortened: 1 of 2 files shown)


DennisPeriquet:
we are not going to go this way.

PR #536, opened 2025-01-24
Support nested parameter (e.g., DEVSCRIPTS_CONFIG) injection from clusterbot

DennisPeriquet:
This allows injecting values for the `DEVSCRIPTS_CONFIG` variable (used in metal tests).

The values must be separated by a `;` e.g.:

```
workflow-launch cucushift-installer-rehearse-baremetalds-ipi-ovn 4.19 "DEVSCRIPTS_CONFIG=INSTALLER_PROXY=true;IP_STACK=v4;NETWORK_TYPE=OVNKubernetes;DENNIS_FOO=BAR"
```

The key is not limited to `DEVSCRIPTS_CONFIG`, e.g:

```
workflow-launch cucushift-installer-rehearse-baremetalds-ipi-ovn 4.19 "FOO_CONFIG=INSTALLER_PROXY=true;IP_STACK=v4;NETWORK_TYPE=OVNKubernetes;DENNIS_FOO=BAR"
```

Here's an example [prow job](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-metal/1882844664378167296) showing the [variables are properly added](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/release-openshift-origin-installer-launch-metal/1882844664378167296/artifacts/launch/baremetalds-devscripts-setup/artifacts/root/dev-scripts/logs/01_install_requirements-2025-01-24-174550.log) (search for `export DENNIS_FOO=BAR` and see the variables before that).

### DIFF
--- a/pkg/slack/slack.go
+++ b/pkg/slack/slack.go
@@ -311,10 +311,24 @@ func BuildJobParams(params string) (map[string]string, error) {
 	jobParams := make(map[string]string)
 	for _, combinedParam := range splitParams {
 		split := strings.Split(combinedParam, "=")
-		if len(split) != 2 {
+		if len(split) > 2 {
+			// We detected nested parameters so process them.
+			multiParams := strings.Join(split[1:], "=")
+			multiSplit := strings.Split(multiParams, ";")
+			value := multiSplit[0]
+			for _, param := range multiSplit[1:] {
+				variable := strings.Split(param, "=")
+				if len(variable) != 2 {
+					return nil, fmt.Errorf("unable to interpret parameter in `%s`. Each nested parameter must be in the form of KEY=VALUE", param)
+				}
+				value += fmt.Sprintf("\n%s=%s", variable[0], variable[1])
+			}
+			jobParams[split[0]] = value
+		} else if len(split) == 2 {
+			jobParams[split[0]] = parseParameterValue(split[1])
+		} else {
 			return nil, fmt.Errorf("unable to interpret `%s` as a parameter. Please ensure that all parameters are in the form of KEY=VALUE", combinedParam)
 		}
-		jobParams[split[0]] = parseParameterValue(split[1])
 	}
 	return jobParams, nil
 }
--- a/pkg/slack/slack_test.go
+++ b/pkg/slack/slack_test.go
@@ -43,6 +43,36 @@ func TestBuildJobParams(t *testing.T) {
 			expected:    map[string]string{"KEY1": "VALUE1"},
 			errorString: "",
 		},
+		{
+			name:        "SingleNESTED_PARAMETER",
+			params:      "\"NESTED_PARAMETER=KEY1=VALUE1\"",
+			expected:    map[string]string{"NESTED_PARAMETER": "KEY1=VALUE1"},
+			errorString: "",
+		},
+		{
+			name:        "MultipleNESTED_PARAMETERs",
+			params:      "\"NESTED_PARAMETER=KEY1=VALUE1;KEY2=VALUE2\"",
+			expected:    map[string]string{"NESTED_PARAMETER": "KEY1=VALUE1\nKEY2=VALUE2"},
+			errorString: "",
+		},
+		{
+			name:        "MixedParametersWithNESTED_PARAMETERs",
+			params:      "\"BASELINE_CAPABILITY_SET=None\",\"NESTED_PARAMETER=KEY1=VALUE1;KEY2=VALUE2\"",
+			expected:    map[string]string{"BASELINE_CAPABILITY_SET": "None", "NESTED_PARAMETER": "KEY1=VALUE1\nKEY2=VALUE2"},
+			errorString: "",
+		},
+		{
+			name:        "MultipleMixedParameters",
+			params:      "\"BASELINE_CAPABILITY_SET=None\",\"NESTED_PARAMETER=KEY1=VALUE1;KEY2=VALUE2\",\"OTHER_CONFIG=KEY1=VALUE1a;KEY2=VALUE2a\"",
+			expected:    map[string]string{"BASELINE_CAPABILITY_SET": "None", "NESTED_PARAMETER": "KEY1=VALUE1\nKEY2=VALUE2", "OTHER_CONFIG": "KEY1=VALUE1a\nKEY2=VALUE2a"},
+			errorString: "",
+		},
+		{
+			name:        "InvalidMixedParameters",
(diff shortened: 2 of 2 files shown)


DennisPeriquet:
> I see a [unit test](https://github.com/openshift/ci-chat-bot/blob/f439307dce096a1586bb0733ffcb0391e122c992/pkg/slack/parser/commandParser_test.go#L224) that has multiple `key=value` pairs separated by commas. Need to ensure this case is handled with my new logic.

The code can remain as-is because the parameters are already split by the time it gets to my new code. I still added a unit test for this case just to test it out.

pkg/slack/slack.go
```diff
@@ -310,6 +310,18 @@ func BuildJobParams(params string) (map[string]string, error) {
 	}
 	jobParams := make(map[string]string)
 	for _, combinedParam := range splitParams {
+		// Check for DEVSCRIPTS_CONFIG ; if found, we know we're dealing with semicolon separated parameters.
```
bradmwilliams, line 313: I don't think it's really necessary to lock this logic down to only `DEVSCRIPTS_CONFIG`.  I'd much rather it be generic so that if/when another case comes along, we don't have to add another condition to the code.  What's important is the syntax of how to pass these variables into the ClusterBot.

pkg/slack/slack_test.go
```diff
@@ -43,6 +43,24 @@ func TestBuildJobParams(t *testing.T) {
...
+			params:      "\"DEVSCRIPTS_CONFIG=KEY1=VALUE1;KEY2=VALUE2\"",
+			expected:    map[string]string{"DEVSCRIPTS_CONFIG": "KEY1=VALUE1\nKEY2=VALUE2"},
+			errorString: "",
+		},
+		{
+			name:        "OtherParametersWithDEVSCRIPTS_CONFIGParameters",
+			params:      "\"BASELINE_CAPABILITY_SET=None\",\"DEVSCRIPTS_CONFIG=KEY1=VALUE1;KEY2=VALUE2\"",
+			expected:    map[string]string{"BASELINE_CAPABILITY_SET": "None", "DEVSCRIPTS_CONFIG": "KEY1=VALUE1\nKEY2=VALUE2"},
+			errorString: "",
+		},
```
bradmwilliams, line 63: Update these to be generic to match the modified logic from above.  Prefer `Single`, `Multiple`, `Mixed` for naming structures instead of `One`, `Two`, etc.

pkg/slack/slack.go
```diff
@@ -310,6 +310,18 @@ func BuildJobParams(params string) (map[string]string, error) {
...
+		if strings.Contains(combinedParam, "DEVSCRIPTS_CONFIG") {
+			// Split by the equal, treat the first part as the key, treat the rest as the value
+			parts := strings.SplitN(combinedParam, "=", 2)
+			if len(parts) == 2 {
+				nParts := strings.Split(parts[1], ";")
+				jobParams[parts[0]] = strings.Join(nParts, "\n")
```
bradmwilliams, line 319: This logic seems risky for setting environment variables...
What if someone passes in: `VARIABLE=A=B;C;D=E;F`?

bradmwilliams (changes requested):

pkg/slack/slack.go
```diff
@@ -309,7 +309,31 @@ func BuildJobParams(params string) (map[string]string, error) {
...
 	}
 	jobParams := make(map[string]string)
+
+	// A regex to look for a parameter of the form KEY=KEY1=VALUE1
+	multiValuePattern := regexp.MustCompile(`^[^=]+=[^=]+=.+`)
 	for _, combinedParam := range splitParams {
+		// Check for a parameter of the form KEY=KEY1=VALUE1
+		if multiValuePattern.MatchString(combinedParam) {
+			// Split by the equal, treat the first part as the key, treat the rest as the value
+			parts := strings.SplitN(combinedParam, "=", 2)
```
bradmwilliams, line 319: I'm still not convinced that adding this logic is even necessary, since at line 337-338 it pretty much accomplishes the exact same thing...

bradmwilliams (changes requested):

pkg/slack/slack.go
DennisPeriquet: Please clarify here. This line is there to parse out the LHS as the variable and the RHS of the equal sign as the nested parameters.  It differs from the `strings.Split(combinedParam, "=")` line below because it has that optional n=2 parameter because I only want to split on the first equal sign.

bradmwilliams: How is the following logic:
```
if multiValuePattern.MatchString(combinedParam) {
	// Split by the equal, treat the first part as the key, treat the rest as the value
	parts := strings.SplitN(combinedParam, "=", 2)
	if len(parts) == 2 {
```
any different/better than what's already being tested for down below:
```
split := strings.Split(combinedParam, "=")
if len(split) != 2 {
```
if the same amount of processing is still required for either path...
```
nParts := strings.Split(parts[1], ";")
// Ensure each key/value pair has the form KEY=VALUE
var validatedKeyPairs []string
for _, part := range nParts {
	keyValue := strings.Split(part, "=")
	if len(keyValue) != 2 {
		return nil, fmt.Errorf("unable to interpret parameter in `%s`. Each parameter must be in the form of KEY=VALUE", part)
	}
	validatedKeyPairs = append(validatedKeyPairs, part)
}
jobParams[parts[0]] = strings.Join(validatedKeyPairs, "\n")
continue
```
?

Said differently, why introduce the Regex at all?

At the end of the day, I understand and appreciate that this logic does what it set out to do, but it introduces a completely separate path (with potentially expensive complexity) to test for a scenario that it was already testing for (and continues to test for after the changes).

bradmwilliams: Ultimately, my initial test for this was pretty close to what you have here actually!  I only cared about this particular case (no clusterbot needed) and my `split := strings.Split(combinedParam, "=")` matches Line 313...

```
var combinedParam = "DEVSCRIPTS_CONFIG=INSTALLER_PROXY=true;IP_STACK=v4v6;NUM_WORKERS=3;NETWORK_TYPE=OVNKubernetes"

split := strings.Split(combinedParam, "=")

// Handle multiple variables...
if len(split) > 2 {
	multiParams := strings.Join(split[1:], "=")
	multiSplit := strings.Split(multiParams, ";")
	var value string
	for _, param := range multiSplit {
		variable := strings.Split(param, "=")
		if len(variable) != 2 {
			continue
		}
		value += fmt.Sprintf("%s=%s\n", variable[0], variable[1])
	}
	jobParams[split[0]] = value
}
```

DennisPeriquet: ah, now I gotcha -- nice.

bradmwilliams (approved):

PR #537, opened 2025-01-28
update FAQ for nested parameters

DennisPeriquet:


docs/FAQ.md
```diff
@@ -43,6 +43,10 @@
  
    `workflow-launch openshift-e2e-gcp 4.18 "BASELINE_CAPABILITY_SET=None","ADDITIONAL_ENABLED_CAPABILITIES=CloudControllerManager CloudCredential Console Ingress MachineAPI"`
 
+   The parameters can have nested parameters, for example:
+
+   `workflow-launch openshift-e2e-gcp 4.18 "BASELINE_CAPABILITY_SET=None","ADDITIONAL_ENABLED_CAPABILITIES=CloudControllerManager CloudCredential Console Ingress MachineAPI"," "MORE_PARAMETERS=AGENT_E2E_TEST_SCENARIO=HA_IPV4V6_DHCP;MIRROR_IMAGES=true"`
+
```
bradmwilliams, line 49: There appears to be an extra space and quote in your string.

Also, I would recommend explaining what "nested parameters" means and the specific syntax on how to use them.  Perhaps in their own example, not as part of a previous example.

bradmwilliams (changes requested):

bradmwilliams (approved):

PR #547, opened 2025-03-03
OCPCRT-367,OCPCRT-368,OCPCRT-370: MCE Improvements

AlexNPavel:
This PR makes improvements to 3 areas of the MCE support:
- removes unnecessary empty message when calling `mce auth`
- removes requirement of cluster name for `auth` and `delete` commands when user only has 1 active cluster
- adds an annotation to the managed cluster after a user is notified to prevent extra notifications after the first

Depends on https://github.com/openshift/continuous-release-jobs/pull/1574


bradmwilliams (approved):
Great improvements!  Thanks!!

PR #554, opened 2025-04-09
Update linting and modernize

AlexNPavel:
Fix all issues listed by golangci-lint's latest updates and run the `gopls` modernizer.

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -18,12 +18,12 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"maps"
 	"os"
 	"path/filepath"
 	"time"
 
 	"github.com/adrg/xdg"
-	"github.com/sirupsen/logrus"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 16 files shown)


pkg/catalog/catalog.go
```diff
@@ -259,7 +257,7 @@ func CreateContent(ctx context.Context, secretAuth []byte, bundleImage string) (
 
 // generateFBCContent creates a File-Based Catalog using the bundle image and context
 func generateFBCContent(ctx context.Context, f *fbcutil.FBCContext, bundleImage string) (string, error) {
-	log.Infof("Creating a File-Based Catalog of the bundle %q", bundleImage)
+	logrus.Infof("Creating a File-Based Catalog of the bundle %q", bundleImage)
```
bradmwilliams, line 260: Why bother with `logrus` at all here?  We're using `klog` everywhere else in this same file

pkg/catalog/catalog.go
```diff
@@ -278,7 +276,7 @@ func generateFBCContent(ctx context.Context, f *fbcutil.FBCContext, bundleImage
 		return "", fmt.Errorf("error validating and converting the declarative config object to a string format: %v", err)
 	}
 
-	log.Infof("Generated a valid File-Based Catalog")
+	logrus.Infof("Generated a valid File-Based Catalog")
```
bradmwilliams, line 279: Same as above

bradmwilliams (changes requested):
Looks good.  Just one little nit really...

bradmwilliams (approved):

PR #565, opened 2025-08-05
Add rhcos-devel pattern for images

DennisPeriquet:


pkg/manager/manager.go
```diff
@@ -1924,6 +1924,9 @@ var validVersionRegexes = []*regexp.Regexp{
 
 	// Releases built with clusterbot 'build' command
 	regexp.MustCompile(`^registry\.build.*\.ci\.openshift\.org/ci.*/release`),
+
+	// General registry.ci.openshift.org PullSpecs (with tag or SHA)
+	regexp.MustCompile(`^registry\.ci\.openshift\.org/[^/]+/[^/:@]+(?::[^@]+|@sha256:[a-f0-9]{64})`),
```
bradmwilliams, line 1929: A few concerns:
* `[^/]+/` - is incredible broad and can contain characters that are not actually supported by kubernetes namespaces (i.e. @#$%^&)
* `@sha256:[a-f0-9]{64})` - base64 strings are case sensitive, so this won't actually work for everything.

bradmwilliams (changes requested):

pkg/manager/manager.go
bradmwilliams: There might be a library somewhere the `containers` organization that handles this already... maybe? 
Might be worth a look!

DennisPeriquet: thanks! I tried to cover all that the grammar allows -- I'll admit the regex is rather long now

bradmwilliams: > rather long now

That's an under statement! :laughing: 

My comment above, about `reference.go`, was more about leveraging the library than to re-implement similar logic...
https://gist.github.com/bradmwilliams/247c4073c2628cf351db1a9ac0cd8adf (AI Generated)

pkg/manager/manager.go
```diff
@@ -1890,6 +1891,12 @@ func (m *jobManager) CheckValidJobConfiguration(req *JobRequest) error {
 	return nil
 }
 
+var allowedRegistries = []string{
+	"registry.ci.openshift.org",
+	"quay.io",
+	"registry.build06.ci.openshift.org",
```
bradmwilliams, line 1897: There are 11 possible build clusters now...
https://docs.ci.openshift.org/docs/getting-started/useful-links/#clusters

bradmwilliams (changes requested):

pkg/manager/manager.go
DennisPeriquet: yup, then we don't have to fetch the build* ones dynamically. I split it up by static name and regex matching patterns.

pkg/manager/manager_test.go
```diff
@@ -96,6 +96,41 @@ func Test_containsValidVersion(t *testing.T) {
...
+			},
+			want: false,
+		},
+		{
+			name: "Invalid registry.ci PullSpec (missing tag but has v prefix): 'launch registry.ci.openshift.org/rhcos-devel/v4.20.0'",
+			args: args{
+				listOfImageOrVersionOrPRs: []string{"registry.ci.openshift.org/rhcos-devel/v4.20.0"},
+			},
+			want: false,
+		},
```
bradmwilliams, line 133: I don't see a test that validates an item in the `allowedRegistryNames` list and I'd also list to see at least one of these new tests pointing to a registry other than `registry.ci.openshift.org` (just for coverage that the regex works for something more than the 1 registry).

bradmwilliams (changes requested):

bradmwilliams (approved):

PR #572, opened 2025-09-17
manager: add launch label to custom imagesets

AlexNPavel:
When multiple cluster bots are running, they will delete each others imagesets. This PR adds the LaunchLabel to the imagesets so the Cluster Bot can identify which imagesets it owns.


bradmwilliams (approved):
/hold
It's late.  /unhold when you're ready!

PR #604, opened 2026-03-17
OCPCRT-450: Improve ClusterBot's cloud platform dispursment

thiagoalessio:
## Background — approach changed mid-flight

This PR originally reworked ClusterBot's cloud-account balancing by consolidating the per-account Boskos logic into a single `platformQuotaSlices` map and a generic `selectCloudAccountProfile` that queried Boskos metrics and picked the account with the most free leases.

That approach has been **superseded by a Test Platform change** (see [Slack thread](https://redhat-internal.slack.com/archives/CNHC2DK2M/p1788875782782009) and [OCPCRT-450](https://redhat.atlassian.net/browse/OCPCRT-450) comments):

- The "regular" per-account cluster profiles for aws/gcp/azure were replaced by cluster-profile **sets** — `openshift-org-aws`, `openshift-org-gcp`, `openshift-org-azure` — which randomly select an underlying account **at job runtime**. Test Platform now does the account dispersement server-side, which is exactly what this PR set out to do in ClusterBot.
- As part of that migration, all `gcp-*-quota-slice` Boskos resource types were deleted, causing the reported launch failure `failed to get metrics for gcp leases: status 404 Not Found`. The aws/azure quota-slices are on the same path.

So the Boskos-querying approach is now both broken (404 on GCP) and redundant.

## What this PR does now

Delegates account dispersement to Test Platform by launching amd64 clusters under the platform's profile set, instead of querying Boskos:

- Replaces `platformQuotaSlices` / `selectCloudAccountProfile` with a static `platformProfileSets` map (`aws→openshift-org-aws`, `gcp→openshift-org-gcp`, `azure→openshift-org-azure`).
- Replaces `Job.CloudAccountProfile` with `Job.CloudProfileSet`.
- Simplifies `applyClusterProfile` to set only the `cloud-cluster-profile` label and the launch test's `ClusterProfile`; it no longer overrides the per-account secret volume or `BASE_DOMAIN`, since the runtime resolves those from the account the set selects.

## Deferred to a follow-up

The now-unused Boskos lease-client plumbing (`LeaseClient`, `jobManager.lClient`, `NewJobManager` param, `initializeLeaseClient`) and the `--lease-server` flags / lease-credentials secret are left wired to keep this change reviewable and to avoid touching the deployment config in lockstep. They'll be removed once profile-set behaviour is confirmed in practice.

## To confirm with Test Platform

- That a profile set makes ci-operator resolve the per-account secret and `BASE_DOMAIN` automatically (this PR assumes yes and stops overriding both).
- That all workflows ClusterBot launches work under the set profiles.
- Whether the `amd64`-only gate should stay (non-amd64 currently keeps the default per-platform profile).

## How to review

The earlier commits build up the (now-superseded) Boskos refactor; the final commit `Delegate account dispersement to cluster-profile sets` performs the pivot. Reviewing that last commit on its own gives the clearest picture of the end state.

## Summary by CodeRabbit

* **New Features**
  * AMD64 jobs now use the appropriate Test Platform cluster-profile set for AWS, Azure, or GCP.
  * Jobs on other architectures continue using the default cloud profile behavior.

* **Refactor**
  * Cloud account selection is now handled through platform cluster profiles rather than lease-based quota metrics.
  * Simplified cloud profile application while preserving existing cluster-profile labels and launch configuration.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -2229,44 +2229,11 @@ func (m *jobManager) LaunchJobForUser(req *JobRequest) (string, error) {
 
 	// check what leases are available for platform
 	if req.Architecture == "amd64" && m.lClient != nil {
-		switch req.Platform {
-		case "aws":
-			metrics1, err := m.lClient.Metrics("aws-quota-slice")
-			if err != nil {
-				return "", fmt.Errorf("failed to get metrics for `aws` leases: %v", err)
-			}
-			metrics2, err := m.lClient.Metrics("aws-2-quota-slice")
-			if err != nil {
-				return "", fmt.Errorf("failed to get metrics for `aws-2` leases: %v", err)
-			}
-			if metrics2.Free > metrics1.Free {
-				job.UseSecondaryAccount = true
-			}
-		case "azure":
-			metrics1, err := m.lClient.Metrics("azure4-quota-slice")
-			if err != nil {
-				return "", fmt.Errorf("failed to get metrics for `azure` leases: %v", err)
-			}
-			metrics2, err := m.lClient.Metrics("azure-2-quota-slice")
-			if err != nil {
-				return "", fmt.Errorf("failed to get metrics for `azure-2` leases: %v", err)
-			}
-			if metrics2.Free > metrics1.Free {
-				job.UseSecondaryAccount = true
-			}
-		case "gcp":
-			metrics1, err := m.lClient.Metrics("gcp-quota-slice")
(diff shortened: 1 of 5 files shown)


bradmwilliams (changes requested):
/approve

bradmwilliams (approved):
/hold
It's been a bit.  Just want to verify that this is still good to go!  Unhold if so

bradmwilliams (approved):

PR #630, opened 2026-07-23
OCPCRT-525: Adding support for CI Images from QCI

bradmwilliams:
When users attempt to launch/build clusters from a CI release, that points to a reference release in QCI, the clusterbot is unable to resolve the `tag.Image` value because it is not set on reference releases.

Reference job: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-aws-modern/2080191982801522688

Error message:
```
* could not run steps: step [release:latest] failed: unable to import latest release image: unable to import tag ci-ln-v40f022/release:latest at import (0): ImageStreamImport.image.openshift.io "release" is invalid: spec.images[0].from.name: Invalid value: "registry.ci.openshift.org/ocp/release-5:": invalid reference format 
```

The YAML setting that causes the error is:
```
- name: RELEASE_IMAGE_LATEST
        value: 'registry.ci.openshift.org/ocp/release-5:'
```

This PR adds a check for the empty `tag.Image` value and sets the pullSpec to the `tag.DockerImageReference` instead.

Test build/launch: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-aws-modern/2080278580536807424

rh-pre-commit.version: 2.4.0
rh-pre-commit.check-secrets: ENABLED

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CI image resolution for QCI-style tag events.
  * Correctly handles tag events with an empty image field by using the available Docker image reference.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -1379,9 +1379,16 @@ func (m *jobManager) ResolveImageOrVersion(imageOrVersion, defaultImageOrVersion
 		}
 
 		if tag, name := findImageStatusTag(is, unresolved); tag != nil {
-			klog.Infof("Resolved %s to image %s", imageOrVersion, tag.Image)
+			// CI releases are now all references to QCI, which means they don't have a value for the "Image" field
+			var installSpec string
+			if len(tag.Image) == 0 {
+				installSpec = tag.DockerImageReference
+				klog.Infof("Resolved %s to QCI image %s", imageOrVersion, installSpec)
+			} else {
+				klog.Infof("Resolved %s to image %q", imageOrVersion, tag.Image)
+				installSpec = buildPullSpec(ns, tag.Image, isName)
+			}
 			// identify nightly stream for runspec if not amd64
-			installSpec := buildPullSpec(ns, tag.Image, isName)
 			runSpec := ""
 			if architecture == "amd64" || architecture == "multi" || strings.Contains(unresolved, "konflux") {
 				runSpec = installSpec


hoxhaeris:
I was looking at the same issue this morning. I'm not 100% sure on my analysis here so feel free to correct me, but I noticed something in the same area that might be worth considering.

The `DockerImageReference` fallback looks right for the primary amd64 path. However, the same empty-`Image` pattern exists at [line 1399-1400](https://github.com/openshift/ci-chat-bot/blob/67522881/pkg/manager/manager.go#L1399-L1400), which handles arm64 non-nightly launches:

```go
runTag, _ := findImageStatusTag(amd64IS, unresolved)
runSpec = buildPullSpec("ocp", runTag.Image, "release")
```

If I'm reading this correctly, this has two potential issues:
1. **Same empty-Image case** — `runTag.Image` could also be empty for QCI references in the amd64 imagestream, producing the same invalid reference
2. **Nil pointer risk** — if the tag isn't found in the amd64 imagestream, `runTag` is nil and `runTag.Image` would panic

The `findSpecTagByName` path at [line 1438](https://github.com/openshift/ci-chat-bot/blob/67522881/pkg/manager/manager.go#L1438) handles this more defensively — it checks `runTag != nil` and falls back to `installSpec`. Something similar here might be worth adding:

```go
} else {
    runTag, _ := findImageStatusTag(amd64IS, unresolved)
    if runTag != nil && len(runTag.Image) > 0 {
        runSpec = buildPullSpec("ocp", runTag.Image, "release")
    } else if runTag != nil && len(runTag.DockerImageReference) > 0 {
        runSpec = runTag.DockerImageReference
    } else {
        runSpec = installSpec
    }
}
```

hoxhaeris (approved):

